PR #119, opened 2014-09-14
Attempt at resolving how to render chapter cues.

silviapfeiffer:
This patch is trying to fill the gap in
http://dev.w3.org/html5/webvtt/#cues-in-isolation

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -3659,13 +3659,31 @@ The Final Minute</pre>
 
     <ol>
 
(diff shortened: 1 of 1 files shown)


foolip:
Finally I get around to looking at the now-resolved HTML bug, and this is what I propose:
https://github.com/w3c/webvtt/pull/172

silviapfeiffer:
@foolip had a better patch

PR #171, opened 2015-01-29
Simplify snap-to-lines positioning to discard cues that don't fit

foolip:
Instead of picking a "best position" which can both overlap other cues
and have lines outside the title area, simply discard cues when no
non-overlapping and contained position is found.

This achieves the "Don't overlap cues that don't fit" goal from
https://www.w3.org/Bugs/Public/show_bug.cgi?id=17483

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -4382,13 +4382,6 @@ The Final Minute</pre>
           <li><p>Remember the position of all the boxes in <var>boxes</var> as their <var>specified
           position</var>.</p></li>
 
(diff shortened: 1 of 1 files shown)


silviapfeiffer:
Ready to be accepted in a week if there are no new comments as per https://www.w3.org/Bugs/Public/show_bug.cgi?id=17483#c27 .

foolip:
Sounds good to me.

foolip:
Thanks!

silviapfeiffer:
Sorry for the delay. Catching up on things!

foolip:
No problem, it seemed like a week to me :)

PR #172, opened 2015-02-24
Define the WebVTT rules for extracting the chapter title

foolip:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=27102

This doesn't actually define the WebVTT rules, that is for later.


foolip:
I also throw a move commit in there. Use `git show -w --word-diff` to confirm that there's nothing else in there.

foolip:
OK, I went the whole way in the PR, renamed to reflect that.

foolip:
Ping

silviapfeiffer:
Sorry about the delay - this needs a look from the command-line and haven't had time yet. Doing it now...

silviapfeiffer:
Nice!

foolip:
Thanks for the review!

silviapfeiffer:
Any time! (well... when I have the headspace :-)

PR #173, opened 2015-02-25
Rename the section on WebVTT cues

foolip:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28070

This leaves the "Text track regions" section looking out of place:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28098


silviapfeiffer:
See also https://github.com/w3c/webvtt/pull/177

PR #174, opened 2015-02-25
Remove inaccurate description of writing direction

foolip:
This paragraph appears to conflate the writing direction, which is a
setting on the cue, with the paragraph embedding level and direction,
which are properties determined by applying the Unicode Bidirectional
Algorithm.

Introduced in commit 86ba250eaf7d2e53f614d684755ef13d089c7652.

Noticed due to https://www.w3.org/Bugs/Public/show_bug.cgi?id=28071


silviapfeiffer:
Fair enough.

PR #175, opened 2015-02-25
Don't use quotes around 'true' as it is not a string

foolip:
This is consistent with many other uses of "be true" and "be false",
as in "Let cue's text track cue snap-to-lines flag be true."


silviapfeiffer:
Was that really the only place in the whole doc with quotes on boolean values?

foolip:
Yep, just the one :)

PR #177, opened 2015-03-08
Move 'WebVTT cue' def to Data Model and use consistently.

silviapfeiffer:
'WebVTT cue' is now a concept and can thus replace the use of 'text
track cue'. This commit only does it where 'text track cue' is not used
with any other property, e.g. 'text track cue identifier' for
readability reasons.

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -698,15 +698,15 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
    <section>
     <h3>WebVTT cues</h3>
 
-    <p>WebVTT cues are HTML <a title="text track cue">text track cues</a> that additionally consist
-    of the following: [[!HTML]]</p>
+    <p>A <dfn>WebVTT cue</dfn> is a HTML <a title>text track cue</a> that additionally consist of
+    the following: [[!HTML]]</p>
 
     <dl>
 
      <dt><dfn title="text track cue box">A cue box</dfn></dt>
      <dd>
-      <p>The cue box of a <a>text track cue</a> is a box within which the text of all lines of the
-      cue is to be rendered.</p>
+      <p>The cue box of a <a>WebVTT cue</a> is a box within which the text of all lines of the cue
+      is to be rendered.</p>
 
       <p class="note">The position of the <a title="text track cue box">cue box</a> within the video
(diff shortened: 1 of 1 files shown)


webvtt.html
```diff
@@ -1171,7 +1168,7 @@
     contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E
     GREATER-THAN SIGN).</p>
 
-    <p>A <dfn>WebVTT cue</dfn> consists of the following components, in the given order:</p>
```
foolip: This concept is needed, it's referenced from "WebVTT file body" which is now pointing to the wrong definition.

foolip:
This is good, but we need to move the existing definition out of the way first:
https://github.com/w3c/webvtt/pull/180

webvtt.html
silviapfeiffer: Looking at the WHATWG HTML spec, Ian also always links back to the "text track cue" even when in the parsing sections of the video element. He doesn't distinguish between the concept and its instantiation. Should we really be separating the two?

foolip: Text track cues don't have a syntax, which parsing do you mean?

silviapfeiffer: E.g. this paragraph:
"Each media element has a list of newly introduced cues, which must be initially empty. Whenever a text track cue is added to the list of cues of a text track that is in the list of text tracks for a media element, that cue must be added to the media element's list of newly introduced cues. Whenever a text track is added to the list of text tracks for a media element, all of the cues in that text track's list of cues must be added to the media element's list of newly introduced cues. When a media element's list of newly introduced cues has new cues added while the media element's show poster flag is not set, then the user agent must run the time marches on steps."

But I can live with "webvtt cue box". I'll rebase this and then we can try again.

foolip: To elaborate, "Zero or more WebVTT cues and WebVTT comments separated from each other by one or more WebVTT line terminators." would be nonsensical if "WebVTT cue" was a concept and not a piece of syntax.

foolip:
Ugh, I merged this manually to resolve the conflicts before I saw that you had updated the branch. I've confirmed that merging your updated branch would have given identical results, so closing this.

silviapfeiffer:
OK :-)

PR #178, opened 2015-03-08
Use consistent linking scheme to whatwg spec.

silviapfeiffer:
All other links got directly to the single page spec and only this one
was to the multipage version.


foolip:
Oops!

silviapfeiffer:
it's what peer review is for. :-)

PR #179, opened 2015-03-08
Add a diagram describing WebVTT region anchoring.

silviapfeiffer:
Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28075


webvtt.html
```diff
@@ -1094,6 +1094,12 @@
      </dd>
     </dl>
 
+    <p class="note">The following diagram illustrates how anchoring of a region to a video viewport
+    works. The black cross is the anchor, orange explains the anchor's offset within the region and
+    green the anchor's offset within the viewport. Think of it as sticking a pin through a note onto
+    a board:</p>
+    <img src="webvtt_region_diagram.png" alt="region achoring displayed visually"></img>
```
foolip: s/achoring/anchoring/

silviapfeiffer:
Good catch, thanks!

foolip:
The diagram says "WebVTT Cue Box" for the outermost box, should be...? There's a bug for that: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28105

foolip:
I'd suggest "rendering area" as in "The output of the steps below is a set of CSS boxes that covers the rendering area of the media element or other playback mechanism, which user agents are expected to render in a manner suiting the user." but that term is overloaded in "A text track region represents a subpart of the video viewport and provides a rendering area for text track cues."

Going with "video viewport" on the other hand presumes a certain answer to that spec bug.

silviapfeiffer:
We are also using "video viewport" for this. Might be worth clarifying that the "Cue Box" is the video viewport?

silviapfeiffer:
I agree with Fredrik and we haven't had any other advice. Maybe set a deadline for feedback on that bug?

silviapfeiffer:
Actually, I am confusing myself. The viewport is the whole dimension of the video, while the cue box is supposed to only related to the dimensions of the cue. I will change the image and this is unrelated to the bug.

foolip:
OK, it now says "video viewport" which is consistent with how regions are described elsewhere. Maybe we'll have to mass rename later.

I'll merge this in a bit, it will require manually adding the image when publishing I think.

silviapfeiffer:
Yeah, I think you might need to run 'cvs add' at some point by hand...

foolip:
Now merged with some minor changes, see the merge commit. (Not separately as that would have resulted in both files checked into CVS.)

PR #180, opened 2015-03-09
Rename the "WebVTT cue" syntax to "WebVTT cue block"

foolip:
For consistency also rename "WebVTT comment" to "WebVTT comment block",
since that is also a multi-line construct.

Also merge the WebVTT comments section into the file structure section,
since no other bits of syntax get their own section.


silviapfeiffer:
I guess that works...

PR #184, opened 2015-03-09
Rename "text track cue/region *" to "WebVTT cue/region *"

foolip:


foolip:
This will be no fun at all to review, even `git diff --word-diff -w` is pretty big. I used a script for each step and then combined some steps. Maybe it's easier to review the script:
```sh
#!/bin/sh -e

./format.py webvtt.html 100000
sed "s/$1/$2/g" < webvtt.html > tmp.html
mv tmp.html webvtt.html
./format.py webvtt.html 100
git add webvtt.html
git commit -m "Rename \"$1\" to \"$2\""
```

This also requires `COLUMNS = int(sys.argv[2])` in format.py. As a result I couldn't separate the reflow from the rename, so lots of lines have changed.

The very last commit is worth looking at manually I think.

Note that this leaves us in a pretty ugly state where "text track region" still appears in a few places that weren't regular enough for my script and it's now apparent that "text track cue text" needs to be defined by us as "WebVTT cue text", which unfortunately will conflict.

Just let me know which bits you'd like to do here and which can be separately.

foolip:
This will make pretty much any other change conflict, so merging the good bits fast would be good.

silviapfeiffer:
Wasn't that hard to do - github shows the diffs pretty nicely. I still think you need to add that dfn markup in.

PR #185, opened 2015-03-11
Add support for STYLE blocks (inline style)

foolip:
This is Simon Pieter's proposal with only minor typo fixes:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=15023#c31

### DIFF
--- a/config-cg.js
+++ b/config-cg.js
@@ -54,12 +54,26 @@ var respecConfig = {
             status: "ED",
             title: "CSS Color Module Level 4",
         },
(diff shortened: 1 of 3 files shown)


foolip:
Please do not merge this yet, I will post to the spec bug about some things I think should change.

plehegar:
See bug https://www.w3.org/Bugs/Public/show_bug.cgi?id=15023

foolip:
@zcorpan, feel free to close this when you have a new pull request for it.

PR #186, opened 2015-03-17
Don't automatically add *.png to CVS

foolip:
cvs add fails if the file is already added, so add new images manually
when needed, which will be rarely.


foolip:
This had broken the webhook and prevented the spec from being updated, so merged without review.

silviapfeiffer:
No worries :-)

PR #187, opened 2015-03-17
Drop the WebVTT prefix from cue computed definitions

foolip:
It's very repetetive in cases like "A WebVTT cue has a WebVTT cue
computed line position" and isn't needed to disambiguate.

This is the first commit of several indented to separate the
defintitions as follows:

 * "cue x" for the text track cue and WebVTT model
 * "WebVTT cue x" for the syntax
 * "vttcue-x" for the VTTCue API


foolip:
PTAL. I'm trying to get things into a state where we don't have the "text track" and "WebVTT" prefixes everywhere so that it reads more like English, and also no naming conflicts between the syntax and the model. I ran into this again when trying to define "WebVTT cue text", since "text track cue text" is gone from HTML :(

silviapfeiffer:
Agreed, this is the WebVTT spec, we don't need to repeat WebVTT in all the definitions.

PR #188, opened 2015-03-22
The word "contribution" has legal implications.

silviapfeiffer:
The legal implications of "contribution" on IPR are far more involved
than what the actual contributions of the listed people were, in
particular with respect to patents. Avoid the word.

As requested in an email thread with Thierry, Philippe and David.

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -5574,8 +5574,8 @@ interface <dfn>VTTRegion</dfn> {
    the SubRip software program whose SRT file format was used as the basis for the WebVTT text track
    file format.</p>
 
-   <p>Thanks to the many contributors to the HTML standard, where WebVTT was originally specified.
-   [[!HTML]]</p>
+   <p>Thanks to Ian Hickson and many others for their work on the HTML standard, where WebVTT was
+   originally specified. [[!HTML]]</p>
 
    <p>
     Further thanks to:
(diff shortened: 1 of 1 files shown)


foolip:
Shouldn't the use just above also change? When written, it was surely meant in the same sense as this instance.

foolip:
Is there no clarification we can make to keep using the existing terminology? GitHub uses the word contributor in this non-legalistic sense: https://github.com/w3c/webvtt/graphs/contributors

We could also copy HTML's format with "Thanks to ... , for their useful comments, both large and small, that have led to changes to this specification over the years."

nigelmegitt:
There are two independent issues here:

1. How contributions are acknowledged in the document;
2. Whether any contributor's contributions are substantive enough to warrant IPR inclusion.

Regardless of the resolution to the first issue, if the contribution may be subject to IPR then the contributor needs to clear it for the Recommendation track document, either explicitly through the W3C FSA process or by agreeing to the terms of membership of the relevant Working Group.

silviapfeiffer:
The wording of the HTML spec is fine by me, too.
David & Nigel may want to comment.

foolip:
For the first instance of "contributor", how about "Thanks to Ian Hickson and many others for their work on the HTML standard, where WebVTT was originally specified."?

That's the best I can come up with that avoid the need to have a noun for the person doing the contributing.

silviapfeiffer:
OK, I'll adjust the PR with that wording.

PR #191, opened 2015-06-07
Improve explanation when not to run the rendering algorithm.

silviapfeiffer:
We could remove that sentence completely, since it's a surplus explanation. But I am aware that many people ask the question why not to pursue with an audio element, so it might be better to improve the sentence based on discussion at https://www.w3.org/Bugs/Public/show_bug.cgi?id=28511

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -3688,11 +3688,11 @@ The Final Minute</pre>
   <section>
    <h2>Rendering</h2>
 
-   <p class="note">This section contains an algorithm that describes in some detail how to visually
-   render the content of WebVTT cues in a user agent. It also specifies extra CSS functionality made
-   available to manipulate the visual rendering. The processing model is quite tightly linked to
-   media elements in HTML. When supporting WebVTT in media players that don't support CSS,
-   equivalent visual rendering will need to be implemented.</p>
+   <p class="note">This section describes in some detail how to visually render WebVTT cues in a
+   user agent. It also specifies extra CSS functionality made available to manipulate the visual
+   rendering. The processing model is quite tightly linked to media elements in HTML. When
+   supporting WebVTT in media players that don't support CSS, equivalent visual rendering will need
+   to be implemented.</p>
 
    <section>
     <h4>Processing model</h4>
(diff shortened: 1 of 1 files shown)


foolip:
Does changing the wording really address https://www.w3.org/Bugs/Public/show_bug.cgi?id=28511? I think what John wants is the ability to use WebVTT with `<audio>`, not different wording for why it doesn't work.

foolip:
That being said, I'm fine with this change, and it could be even less direct about the reasons by just saying "abort these steps" with no explanation.

silviapfeiffer:
I agree that John wants WebVTT to render on <audio>, but that's limited in the HTML spec and not in the WebVTT spec, so we can't give him that. It's not that WebVTT doesn't work with <audio> - it does the parsing and fills the VTTCue fields. It's only the rendering that isn't done.

foolip:
Well, OK. I think this change makes things neither better nor worse, but if you think it'll help us move forward please go ahead and merge.

silviapfeiffer:
I'll wait until John's feedback. I made another suggestion to rename "Rendering" into "Native Browser Rendering" (or something like it), so that might be a better resolution.

silviapfeiffer:
I've not renamed the section, but instead added some non-normative notes to better explain what's happening. PTAL.

webvtt.html
```diff
@@ -3688,6 +3688,12 @@
   <section>
    <h2>Rendering</h2>
 
+   <p class="note">This section contains an algorithm that describes in some detail how to visually
```
foolip: I think we can drop "contains an algorithm that" and "the content of" to make this a bit tighter.

webvtt.html
```diff
@@ -3697,6 +3703,11 @@
     track">text tracks</a> that use these rules for a given <a>media element</a>, or other playback
     mechanism, are rendered together, to avoid overlapping subtitles from multiple tracks.</p>
 
+    <p class="note">Note that in HTML, audio elements don't have a visual rendering area and
```
foolip: Drop the leading "Note that". Also, it's not really true that audio elements don't have a visual rendering area, you can make them any size you like with CSS, it's just that we've decided not to use that area. Perhaps put this more matter-of-factly, like "No cues will be rendered for audio elements..." or similar?

foolip:
Looks good, just a few nits.

silviapfeiffer:
Thanks, fixed. PTAL.

foolip:
Thanks!

PR #194, opened 2015-06-07
Rename "line position" to "line offset".

silviapfeiffer:
Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28073

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -709,7 +709,7 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
 
       <p class="note">The position of the <a title="WebVTT cue box">cue box</a> within the video
       frame's dimensions depends on the value of the <a>WebVTT cue text position</a> and the
-      <a>WebVTT cue line position</a>.</p>
+      <a>WebVTT cue line offset</a>.</p>
 
       <p class="note">Lines are wrapped within the <a title="WebVTT cue box">cue box</a>'s <a
       title="WebVTT cue size">size</a> if lines' lengths make this necessary.</p>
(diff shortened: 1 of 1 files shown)


foolip:
To make this easier to review, is there any part of this that isn't simply replacing "line position" with "line offset" wherever it occurs? The good thing about this change is that it makes it much easier to guess which of "line offset" and "text position" corresponds to `VTTCue.line` and `VTTCue.position` respectively. However, these two concepts are positioning the cue on one axis each, so "position" feels equally valid of invalid in both contexts. If you don't like https://www.w3.org/Bugs/Public/show_bug.cgi?id=28073#c3 then I agree this is better than the current state.

silviapfeiffer:
I've replaced "line position" with "line offset" and also made some changes where "line offset" ended up being described as "positioning the cue box" into "offsetting the cue box".

It indeed makes the document more readable, in particular removes the confusion between line position and text position. I don't like "a line" and "a position", because a line inherently isn't about the position, but about a string of text in my mind. So, "line offset" seems more appropriate.

foolip:
OK, let's go with this and see if we're still confused by the new wording :)

nigelmegitt:
I don't think this change helps much, if at all. Two main concerns: 

1. the suggestion that 'offset' means something perpendicular to 'position' is too subtle and doesn't come from any normal usage of English.
2. the term 'offset' normally refers to a relative change in position, i.e. you need an absolute baseline position relative to which you're adding or subtracting something. In this case there is no absolute position. Sentences like "The cue box's top side ... is aligned at the line offset" are very confusing and hard to read because that's a direct misuse of "offset" in this context.

It also doesn't resolve the basic difficulty: sentences like "The text position defines the offset of the cue box..." demonstrate - are the two concepts of position and offset actually interchangeable, or orthogonal?

I've just noticed that offset is also used to describe times elsewhere in the document.

For me, the original confusion came from the terms "text" and "line", and this edit doesn't address that. I think an alternative edit that would have worked here would have been to use longer more expressive names. The points in decreasing order of scope in my mind are:

* It's a *cue* setting
* It's expressed in either the *inline* or the *block* direction
* It's a *position*

So you'd end up with something like "WebVTT Cue Inline Position" and "WebVTT Cue Block Position". I would critique this though, as being somewhat unwieldy. At least it makes clear that to understand the position you have to know the writing direction though, which is the key point.

foolip:
At the very least, the concept underlying `VTTCue.line` should have "line" in the name and that underlying `VTTCue.position` should have "position" in the name. Preferable, they shouldn't have the other word in the name, so no "line position".

silviapfeiffer:
Nigel, thanks for that feedback.

I still think that "offset" can work, because the lines are offset from the top of the video viewport.

Inline and block have a different meaning and don't actually describe positioning from the top or left edge, but instead describe CSS box properties, so these words don't work.

I'll have a go at fixing up the language around "offset" where you identified issues.

nigelmegitt:
Thanks Silvia, those bits of language are just a couple that stood out - I didn't do an exhaustive search for all the places where the edit caused a language problem.

nschonni:
What about using the CSS naming of margin/padding http://www.w3.org/wiki/The_CSS_layout_model_-_boxes_borders_margins_padding

silviapfeiffer:
"line margin" or "line padding" don't really create the right impressions either. I think I'll give "offset" another go.

foolip:
FWIW, I think position is the right word, the only problem with that is that both definitions have "position" in the name making it less clear what VTTCue.position is.

silviapfeiffer:
See https://github.com/w3c/webvtt/pull/197

silviapfeiffer:
To be honest, the change in https://github.com/w3c/webvtt/pull/197 actually makes this a lot more readable and clear, since it forces us to explain what the line offset is offset from, i.e. which edge of the viewport. I'm happy with this wording.

PR #196, opened 2015-06-08
Disallow WebVTT metadata and cue settings zero length. 

silviapfeiffer:
Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28259


foolip:
The title of this PR is from https://github.com/w3c/webvtt/pull/196 and there's a typo "Discallow", can you tweak and force push so that only the right bits get merged here?

foolip:
The actual changes look good, however, feel free to merge yourself after typo-fixing.

PR #197, opened 2015-06-12
Improve the wording around "line offset".

silviapfeiffer:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28073

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -719,7 +719,7 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
 
       <p class="note">The position of the <a title="WebVTT cue box">cue box</a> within the video
       frame's dimensions depends on the value of the <a>WebVTT cue text position</a> and the
-      <a>WebVTT cue line offset</a>.</p>
+      <a>WebVTT cue line</a>.</p>
 
       <p class="note">Lines are wrapped within the <a title="WebVTT cue box">cue box</a>'s <a
       title="WebVTT cue size">size</a> if lines' lengths make this necessary.</p>
@@ -749,14 +749,14 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
       </ul>
 
       <p>If the <a title="WebVTT cue writing direction">writing direction</a> is <a title="WebVTT
-      cue horizontal writing direction">horizontal</a>, then the <a title="WebVTT cue line
-      offset">line offset</a> percentages are relative to the height of the video, and <a
-      title="WebVTT cue text position">text position</a> and <a title="WebVTT cue size">size</a>
-      percentages are relative to the width of the video.</p>
+      cue horizontal writing direction">horizontal</a>, then the <a title="WebVTT cue line">cue
+      line</a> percentages are relative to the height of the video, and <a title="WebVTT cue text
+      position">text position</a> and <a title="WebVTT cue size">size</a> percentages are relative
+      to the width of the video.</p>
 
-      <p>Otherwise, <a title="WebVTT cue line offset">line offset</a> percentages are relative to
-      the width of the video, and <a title="WebVTT cue text position">text position</a> and <a
-      title="WebVTT cue size">size</a> percentages are relative to the height of the video.</p>
+      <p>Otherwise, <a title="WebVTT cue line">cue line</a> percentages are relative to the width of
+      the video, and <a title="WebVTT cue text position">text position</a> and <a title="WebVTT cue
+      size">size</a> percentages are relative to the height of the video.</p>
 
       <p>The <a title="WebVTT cue writing direction">writing direction</a> defaults to <a
       title="WebVTT cue horizontal writing direction">horizontal</a>.</p>
@@ -766,10 +766,10 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
      <dt><dfn title="WebVTT cue snap-to-lines flag">A snap-to-lines flag</dfn></dt>
      <dd>
 
-      <p>A boolean indicating whether the <a title="WebVTT cue line offset">line offset</a> is an
-      integer number of lines (using the line dimensions of the first line of the cue), or whether
-      it is a percentage of the dimension of the video. The flag is set when lines are counted,
-      unset otherwise.</p>
+      <p>A boolean indicating whether the <a title="WebVTT cue line">cue line</a> is an integer
+      number of lines (using the line dimensions of the first line of the cue), or whether it is a
+      percentage of the dimension of the video. The flag is set when lines are counted, unset
+      otherwise.</p>
 
       <p>Cues whose <a>WebVTT cue snap-to-lines flag</a> is set will be placed within the title-safe
       area on user agents that use overscan. Cues with the flag unset will be offset as requested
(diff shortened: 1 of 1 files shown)


foolip:
I don't know, I'm starting to regret https://github.com/w3c/webvtt/pull/194 and don't really think this is now any clearer. The "line offset" is no more an offset than the "text position", both of them influence either `x-position` and `y-position` in the rendering algorithm and then either the `left` or `top` CSS properties. Using "indent" isn't great since it's something quite different from `text-indent` in CSS, that only indents the first line.

My preference would be to revert https://github.com/w3c/webvtt/pull/194 and resolve the bug as WONTFIX. Perhaps the "Pictures that visualize cue boxes the writing directions and the positioning concepts would be great." bit of Andreas Tai's could help instead.

tairt:
Thanks for the work on this issues! I agree with Silvia that this now much clearer and much more intuitve. The terms are coming much closer to the concepts and the additional text helps.

The best would be to show the text to somebody who haven´t spent much time on this and see if he/she understands the concepts.

silviapfeiffer:
@foolip I did indeed experiment with the use of 'indent " for the position setting direction. I can refer that if that caused confusion. I still think the wording for the line setting direction is much clearer and good to hear that @tairt agrees.

foolip:
So perhaps we should treat the name of things from their descriptions. I regret changing "line position" to "line offset" and think the old "line position" and "text position" (both are positions in the same sense) or just "line" and "position" to match the API would be preferable. However, the definitions themselves can probably be much improved.

Would going back to the old names or "line" and "position" together with improved definitions be OK?

dwsinger:
> On Jun 18, 2015, at 1:23 , Philip Jägenstedt <notifications@github.com> wrote:
> 
> So perhaps we should treat the name of things from their descriptions. I regret changing "line position" to "line offset" and think the old "line position" and "text position" (both are positions in the same sense) or just "line" and "position" to match the API would be preferable. However, the definitions themselves can probably be much improved.
> 
> Would going back to the old names or "line" and "position" together with improved definitions be OK?

Two answers

a) as long as we’re only changing documentation, not what constitutes a valid VTT file, or the parsers
b) as long as Andreas is happy!


David Singer

silviapfeiffer:
I don't want to revert, to be honest, but instead move forward. The previous wording was too confusing - "line position" really didn't work. Also seems to me that Andreas agrees, though it might be good to get some other opinions, too.

I'm trying to find a way forward.

I think we should keep "line" and "position" as the names of the cue settings. That makes this whole change editorial in nature.

Both of these types of positioning are about offsetting the cue box from the edges of the video viewport.
* "position" is about offsetting the cue box in the inline direction (as [Nigel](https://github.com/w3c/webvtt/pull/194#issuecomment-110051562) rightly pointed out) - that's the writing direction in current WebVTT speak
* "line" is about offsetting the cue box in the block direction

Maybe therefore both should be talking about "cue box offset" and the cue settings just refer to the direction?

tairt:
I agree with all of you that the API and what constitutes a valid WebVTT file should not be changed at this point in time because the discussed cue settings have been already widley be implemented. 

Although I think that the recent edit makes it clearer I would be happy with any edit that makes it more intuitive and understandable to a WebVTT spec reader. To work with the term "cue box offset" in the description could be one way.

I really think that it is important that people who are not involved in the spec process of WebVTT or TTML read the sections and give feedback if they makes sense to them.

silviapfeiffer:
I've asked the CG for input.

foolip:
No feedback, but it sounds like we can both live with using just the names "line" and "position" for the concepts?

silviapfeiffer:
Yes. Though how to describe them is the challenge.

foolip:
Maybe illustrations would communicate this most effectively.

silviapfeiffer:
I've given this a go to return to calling "line offset" -> "(cue) line" and from "text position" -> "(cue) position". See if you are happy with the wording now.

tairt:
Thanks for the edit, Silvia. I think it is a good compromise between consistency with the rest of the spec and the accuracy of the terms. It is good that it is made clear by the naming that the concepts apply to the cue (box).

foolip:
OK, eternity has elapsed, so reviewing this now :)

The change from "video frame" to "video viewport" makes the spec more consistent, which is good. It may need to be refined in order to fix https://www.w3.org/Bugs/Public/show_bug.cgi?id=28105

foolip:
Hmm, so now we have other mismatches, that "A cue line" and "A line alignment" are no longer a pair, and that it's "A line alignment" but "WebVTT cue line alignment".

Maybe:
 * A writing direction
 * A snap-to-lines flag
 * A line (currently *A line offset*)
 * A line alignment
 * A position (currently *A text position*)
 * A position alignment (currently *A text position alignment*)
 * A size
 * An alignment (current *A text alignment*)
 * A region

("A cue box" left alone as it's not actually part of the model, it's the output of these inputs.)

However, at this point I'd be just fine with going back to the previous wording, accepting that both "line position" vs. "text position" have "position" in the name and focusing only on the actual descriptions.

silviapfeiffer:
Why are "cue line" and "line alignment" not matching? They both have "line" in the word. The word "cue" in there is just for clarification - in the "line" definition, it's just "line".

I really don't want to go back to "line position" and was the cause of the original bug.

foolip:
It's just "a size", so why not "a line" and "a position"? Cue is implied in all of these.

But again, I don't think all of this renaming is worth our time. That the original state had issues doesn't imply that there exists a better state or that we have found it, and I think we haven't.

nigelmegitt:
@foolip I'm possibly repeating myself but I think there's clearly a better state where we explicitly use the concepts that are intended, even if they have longer or less 'normal' names. The issue is that we want to be able to express inline direction positions and block direction positions using a writing direction independent term, and to use those positions to relate both to cue boxes (regions) and to text within those boxes in a clear way. 

This issue is explored in the writing-mode CR at http://www.w3.org/TR/css-writing-modes-3/ - I propose that we adopt the concepts and terminology from there to make the conceptual alignment clear, especially those described in §6 of that document.

foolip:
A list of proposed new names compared to the existing names as in https://github.com/w3c/webvtt/pull/197#issuecomment-127595195 would be appreciated, but I don't think this is a good use of anyone's time.

nigelmegitt:
Seems like a reasonable request - not sure if/when I'll be able to do that though.

silviapfeiffer:
I know we want to move on, but I appreciate clear language, so happy to keep tweaking the language. Let's just move on here and open new bugs for the tweaking.

silviapfeiffer:
Registered https://www.w3.org/Bugs/Public/show_bug.cgi?id=29047 for @nigelmegitt

nigelmegitt:
Thanks @silviapfeiffer seems like a good way to move forward.

foolip:
Postponing the renaming and focusing on the wording in this issue sounds good to me. When preparing a revert for https://github.com/w3c/webvtt/pull/194 I noticed that I had accidentally merged this issue in https://github.com/w3c/webvtt/pull/197#event-373922083 so I've reverted that.

foolip:
I've created (but not merged) https://github.com/w3c/webvtt/pull/209

Once that's been merged, please rebase this so that only the wording changes rename, and we can tackle the naming issues as a whole in https://www.w3.org/Bugs/Public/show_bug.cgi?id=29047

silviapfeiffer:
I'm confused: I thought we were good to move forward with this merge and then do the wordsmithing on top of that. I was under the impression that the latter patch was an improvement as perceived by everyone (including yourself) since it went back to calling a line a "(cue) line" and a position a "(cue) position". I didn't think the merge was accidental.

silviapfeiffer:
And, ahem, you've actually merged #209 ...

silviapfeiffer:
Sorry, #209 isn't merged, so we can still stop it if we all agree.

foolip:
This was already merged in https://github.com/w3c/webvtt/pull/197#event-373922083 by accident (mid-discussion), as I noticed due to conflicts as I was preparing https://github.com/w3c/webvtt/pull/209

What I would like is to go back to the old names, improve the wording where possible, and then change the names all at once if a better proposal than the previous (line position + text position) comes along.

foolip:
Since the mess is of my making, I'm also willing to sort it out given that we agree on what we should do. I want to go back to the old names (line position + text position) and leave any renaming to https://www.w3.org/Bugs/Public/show_bug.cgi?id=29047

That would leave improvements to the definitions of the terms, which were collateral damage when I reverted.

silviapfeiffer:
I think you're the only one who wants to go back to the old words is you. Please let's not roll back everything - it was in a pretty good state.

foolip:
OK, I have reverted my revert. The spec is now inconsistent and IMHO worse than before https://github.com/w3c/webvtt/pull/194 but it's better that one of us is happy than to have a middle position that nobody wants.

silviapfeiffer:
What's inconsistent? Could you clarify?

foolip:
First of all, sorry that https://github.com/w3c/webvtt/pull/197#issuecomment-129415928 was a bit harsh, and for the frustrating process that would have been avoided had I thought this through when merging https://github.com/w3c/webvtt/pull/194.

Updating the list from https://github.com/w3c/webvtt/pull/197#issuecomment-127595195, this is the naming that I would propose:

 * A writing direction
 * A snap-to-lines flag
 * A line (originally *A line position*, currently *A cue line*)
 * A line alignment
 * A position (originally *A text position*, currently *A cue position*)
 * A position alignment (originally *A text position alignment*, currently *A cue position alignment*)
 * A size
 * An alignment (currently *A text alignment*?)
 * A region

In other words, I don't understand why *cue* is part of some terms and not others. The term that corresponds to `VTTCue.align` is less obvious, it does have some influence over the CSS `text-align` property, but it also affects the *computed cue position*, so maybe just vaguely calling it "an alignment" would be fine.

I will prepare a PR.

foolip:
OK, see https://github.com/w3c/webvtt/pull/211

I don't know what to do about *A text alignment*, it seems to me that the definition of it is actually wrong, it's written as if it only affected the `text-align` property of text within a box, but it also affects the position of the box via *computed cue position*. So I've left that alone for now.

PR #198, opened 2015-06-14
Add numbered character escapes.

silviapfeiffer:
Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=23867

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -3331,7 +3331,7 @@ The Final Minute</pre>
         <p>Jump to the entry that matches the value of <var>c</var>:</p>
 
         <dl>
-         <dt>'x' character</dt>
+         <dt>'x' or 'X' character</dt>
          <dd>
           <p>Append <var>c</var> to <var>buffer</var> and jump to the step labeled <i>next</i>.</p>
          </dd>
@@ -3344,8 +3344,8 @@ The Final Minute</pre>
          <dt>U+003B SEMICOLON character (;)</dt>
          <dd>
           <p>Map <var>buffer</var> to a Unicode code point by interpreting <var>buffer</var> either
-          as a hexadecimal number (starting with x) or a decimal number. If a Unicode code point is
-          found, append the retrieved unicode character to <var>result</var>.</p>
+          as a hexadecimal number (starting with 'x' or 'X') or a decimal number. If a Unicode code
+          point is found, append the retrieved unicode character to <var>result</var>.</p>
 
           <p>Set <var>tokenizer state</var> to the <a>WebVTT data state</a>, and jump to the step
           labeled <i>next</i>.</p>
--- a/webvtt.html
+++ b/webvtt.html
@@ -634,8 +634,13 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
      <li><a href="https://html.spec.whatwg.org/#collect-a-sequence-of-characters"><dfn>Collect a
      sequence of characters</dfn></a></li>
      <li><a href="https://html.spec.whatwg.org/#ascii-digits"><dfn>ASCII digits</dfn></a></li>
+     <li><a href="https://html.spec.whatwg.org/#ascii-hex-digits"><dfn>ASCII hex digits</dfn></a></li>
      <li><a href="https://html.spec.whatwg.org/#alphanumeric-ascii-characters"><dfn>Alphanumeric
      ASCII characters</dfn></a></li>
+     <li><a href="https://html.spec.whatwg.org/#syntax-charref"><dfn>Decimal numeric character
+     reference</dfn></a></li>
+     <li><a href="https://html.spec.whatwg.org/#syntax-charref"><dfn>Hexadecimal numeric character
+     reference</dfn></a></li>
      <li><a href="https://html.spec.whatwg.org/#space-character"><dfn>Space character</dfn></a></li>
      <li><a href="https://html.spec.whatwg.org/#skip-whitespace"><dfn>Skip whitespace</dfn></a></li>
      <li><a href="https://html.spec.whatwg.org/#split-a-string-on-spaces"><dfn>Split a string on
@@ -1392,6 +1397,8 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
       bidirectional formatting character in the text of the cue.</li>
       <li>A <a>WebVTT cue nbsp escape</a>, representing a U+00A0 NO-BREAK SPACE character in the
       text of the cue.</li>
+      <li>A <a>WebVTT cue numeric character reference</a>, representing a <a>Decimal numeric
+      character reference</a> or a <a>Hexadecimal numeric character reference</a>.</li>
 
      </ul>
 
@@ -1511,6 +1518,10 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
         bidirectional formatting character in the text of the cue.</li>
         <li>A <a>WebVTT cue nbsp escape</a>, representing a U+00A0 NO-BREAK SPACE character in the
         text of the cue.</li>
+        <li>A <a>WebVTT cue numeric character reference</a>, representing a <a>Decimal numeric
+        character reference</a> or a <a>Hexadecimal numeric character reference</a> in the text of
+        the cue.</li>
+
        </ul>
 
       </li>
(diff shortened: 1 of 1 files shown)


foolip:
Should we just give up and support exactly the same escapes as HTML? I see the literal string `&quot;` as part of the subtitles in http://www.tv4play.se over Chromecast, and I'm pretty sure that's because it's using WebVTT with fewer supported escapes than the developers guessed.

silviapfeiffer:
Are there any that wouldn't make sense in WebVTT?

foolip:
Not sure, they all just map to some code point so in principle they should all make as much sense as the code points themselves.

HTML sure has a lot:
https://html.spec.whatwg.org/#named-character-references

XML not so much:
http://www.w3.org/TR/xml11/#sec-predefined-ent

From XML we're only missing `&apos;` and `&quot;`.

Maybe we should consider this question separately, I'll review this as it is soonish :)

silviapfeiffer:
Do we have stats on the usage of all those named char references?

I'd be happy to for now just add those two extra named char references.
I have a feeling the others are better covered with the numbered char escapes.

foolip:
No stats, no.

silviapfeiffer:
OK we should probably keep this PR only for the numbered character references as requested in the associated bug.

dwsinger:
> On Jun 19, 2015, at 13:32 , Silvia Pfeiffer <notifications@github.com> wrote:
> 
> OK we should probably keep this PR only for the numbered character references as requested in the associated bug.
> 
> —
> 

It strikes me that we could (not ideally) encourage software to be liberal in what they accept and translate, and authors to be strict in what they make, by mandating support for a limited set that authors are supposed to stay inside, while recommending support for a wider set.

but it’s not important

David Singer

silviapfeiffer:
As much as possible, I'd like to avoid differences between validity and parsing, because it always opens doors for author freedom with a later need to extend validity.

dwsinger:
> On Jun 19, 2015, at 17:31 , Silvia Pfeiffer <notifications@github.com> wrote:
> 
> As much as possible, I'd like to avoid differences between validity and parsing, because it always opens doors for author freedom with a later need to extend validity.

yes, that was the reason for the parenthetical that it’s not a very good idea…

David Singer

silviapfeiffer:
Check again.

webvtt.html
```diff
@@ -3294,6 +3317,7 @@
          </dd>
 
          <dt>U+003C LESS-THAN SIGN (&lt;)</dt>
+
```
foolip: Stray newline?

webvtt.html
```diff
@@ -1423,6 +1429,8 @@
       bidirectional formatting character in the text of the cue.</li>
       <li>A <a>WebVTT cue nbsp escape</a>, representing a U+00A0 NO-BREAK SPACE character in the
       text of the cue.</li>
+      <li>A <a>WebVTT cue numeric character reference</a>, representing a <a>Decimal numeric
```
foolip: No need to capitalize these concepts in the prose, e.g. "entry settings object" is not capitalized.

webvtt.html
```diff
@@ -3311,6 +3335,39 @@
 
        </dd>
 
+       <dt><dfn>WebVTT numeric character state</dfn></dt>
+
+       <dd>
+
+        <p>Jump to the entry that matches the value of <var>c</var>:</p>
```
foolip: It looks like this doesn't include support for the `&#xFEFF` syntax?

foolip:
So, I think we can do this, but it would be nice to have some show of support for implementors before we go ahead. I'll ask on the bug.

webvtt.html
silviapfeiffer: Not really - the other dt elements stay separate from each other, too.

webvtt.html
silviapfeiffer: OK, fixing.

silviapfeiffer:
hmm, did a rebase instead of a merge with master, dumb.

silviapfeiffer:
OK, fixed.

foolip:
There are some important differences with [https://html.spec.whatwg.org/#tokenizing-character-references](HTML's equivalent steps):
 * HTML treats both `x` and `X` as a hex number.
 * HTML has exceptions for 0x00 and 0x80-0x9F, making e.g. `&#x80;` map to €.
 * HTML maps to U+FFFD "if the number is in the range 0xD800 to 0xDFFF or is greater than 0x10FFFF", while WebVTT drops it all if "a Unicode code point is [not] found".

https://ygg01.github.io/xml5_draft/#charref-tokenizer-head (CC @Ygg01 and @annevk) is also worth a read, asking basically the same questions.

What should we do? It'd be nice if character references worked the same in all parsers, so if we can at least make numerical references work the same it's a start.

annevk:
Not allowing lone surrogates would be good. Not sure about the 0x80-0x9F range, I don't think XML has that. Then again, unclear what the lifetime of XML is, so maybe it's better to match HTML.

Ygg01:
The XML 1.1 (and intriguingly XML 1.0), prohibits entire 0x80-0x9F range, except for [0x85](http://www.codetable.net/hex/85) character. `0x85` is newline character in Unicode, and its used in XML 1.1. In XML 1.0 `0x85` char is allowed but isn't used AFAICT.

silviapfeiffer:
Yeah, I think it's better we align with HTML than with XML, seeing as WebVTT is used mainly with HTML5 and seeing as WebVTT is not an XML format.

Really wished we could just reference https://html.spec.whatwg.org/#tokenizing-character-references ...

annevk:
@silviapfeiffer if you outline what changes are needed to HTML to make it reusable, we could probably make that happen in the near future. Perhaps for now you could write a temporary monkey patch in WebVTT with the changes you want to make in HTML?

silviapfeiffer:
I wonder if we just take it all, i.e. not just the numbered character references, instead just reference all of https://html.spec.whatwg.org/#tokenizing-character-references . @foolip what do you think?

foolip:
The more of the parsing that's shared the better I would say, not sure what the limit is. If it's tricky to share a part of the tokenizer, then at least sharing the rules for how to turn the characters between `&` and `;` into a Unicode code point would be nice.

dwsinger:
> On Aug 10, 2015, at 6:57 , Philip Jägenstedt <notifications@github.com> wrote:
> 
> The more of the parsing that's shared the better I would say, not sure what the limit is. If it's tricky to share a part of the tokenizer, then at least sharing the rules for how to turn the characters between & and ; into a Unicode code point would be nice.

agreed.  is there anything forcing us to be different?


David Singer

silviapfeiffer:
Nothing forcing us to be different. There were some original reasons, simplicity and what's appropriate for captions. But it's likely easier to just do the same as in HTML, even if not all implementations will be in browsers.

zcorpan:
Would be nice to get @hsivonen's perspective here, especially with the idea of sharing HTML's entity parsing.

It seems to me that supporting all of HTML's entities makes implementing a WebVTT parser where there's no HTML parser, a much bigger task. Alternatively, having to pull in an HTML parser to parse WebVTT might be annoying.

Do we want to be exactly like HTML? What about consuming an entity with omitted `;`? What about different rules when in an attribute value? (WebVTT doesn't have attributes like HTML, but I suppose things like `<c.foo&amp;bar>` could be considered like an "attribute value".)

annevk:
I would require `;`. I see this mainly as ensuring that the same entity results in the same code point. It doesn't seem like it would necessarily be a big task to implement, other than requiring a rather large lookup table and some special handling for certain numeric entities.

silviapfeiffer:
OK, so right now we require `;` and we use both `x` and `X` to identify the numeric value.

So, what remains is to:
* add the HTML exceptions for 0x00 and 0x80-0x9F .
* map to U+FFFD "if the number is in the range 0xD800 to 0xDFFF or is greater than 0x10FFFF".

Is that it?

annevk:
I thought you also wanted to enable all entity names, such as `&middot;`.

silviapfeiffer:
Right, so add a reference to this, also: https://html.spec.whatwg.org/#named-character-references

annevk:
Yeah, note that you can make changes to HTML too, if that would make reuse of the text easier.

dwsinger:
the Rec-track document should refer to the Rec-track HTML spec.  is there any difference in this section, or can both the CG report and the WG product refer to it?

> On Sep 27, 2015, at 23:36 , Silvia Pfeiffer <notifications@github.com> wrote:
> 
> Right, so add a reference to this, also: https://html.spec.whatwg.org/#named-character-references
> 
> —
> 

David Singer

zcorpan:
I think there is precedent now for rec-track specs referencing whatwg specs, so that might not be necessary.

zcorpan:
I was thinking of referencing [consume a character reference](https://html.spec.whatwg.org/multipage/syntax.html#consume-a-character-reference) for parsing and [character references](https://html.spec.whatwg.org/multipage/syntax.html#syntax-charref) for syntax. Requiring the `;` in parsing would mean changing HTML, which makes that part of the spec more complex and confusing, which it probably doesn't need (it's already complex and confusing). It also seems to me that requiring `;` in parsing either means that code can't be shared, or we're slowing down the HTML parser. How do people feel about matching HTML exactly for parsing (using the "in text" rules everywhere in WebVTT)?

silviapfeiffer:
I would align it with HTML as much as possible. I didn't see a full alignment with HTML possible, but I like the partial alignment that you are suggesting.

PR #201, opened 2015-06-28
Add requirements for region identifier to exist.

silviapfeiffer:
Addresses https://www.w3.org/Bugs/Public/show_bug.cgi?id=22727


foolip:
The [spec bug](https://www.w3.org/Bugs/Public/show_bug.cgi?id=22727) is from before https://github.com/w3c/webvtt/pull/31

To clarify the handling of region IDs, I would instead make the id->region mapping a parser-internal concern, which would remove the "WebVTT region identifier" concept entirely, and a text track would also not need to have a "text track list of regions" since that is no longer exposed in any way.

WDYT?

silviapfeiffer:
How can you make the id->region mapping an internal concern?

foolip:
`VTTCue` has an `attribute VTTRegion? region` and neither the `VTTCue` nor `VTTRegion` interfaces exposes the identifier. The parser is the only place that actually deals with the region identifiers, so it should be possible to maintain the mapping in the parser and then forget about it.

silviapfeiffer:
Hmm I think there's merely a mistake in the definition of `VTTRegion`: the line that reads

"Let region's WebVTT region identifier be the empty string."

should be removed, since `VTTRegion` doesn't actually expose its identifier. So, I think it already does what you suggest .

As for removing "WebVTT Region identifier" - if we do that, we have to add a optional region to the "WebVTT Cue" model so we can maintain the reference. It's one or the other in the model and I can live with how it is right now..

foolip:
Yep, I think that going all the way, removing the "WebVTT Region identifier" concept and making region identifiers a parser-internal concern is the way to go here. The cue model already has [a region](http://dev.w3.org/html5/webvtt/#dfn-webvtt-cue-region) so I think we're almost there.

silviapfeiffer:
Don't we want to allow for regions to be specified via script and cues to be associated with regions via script?

foolip:
Yes, but you can do it simply by assigning `cue.region = new VTTRegion()` or similar, instead of having the indirection of an ID.

zcorpan:
It seems to me it would be useful to be able to style different regions differently with CSS (feedback from Timed Text meeting at TPAC in Sapporo), which suggests the id should be exposed to Selectors at least, and I think in the API as well so you can create a region that can be identified from CSS.

Currently ::cue-region can't take an id.

Also see https://www.w3.org/Bugs/Public/show_bug.cgi?id=28473

zcorpan:
and https://www.w3.org/Bugs/Public/show_bug.cgi?id=28474

silviapfeiffer:
@zcorpan so what's your suggestion? Keep the id? This patch is ok to go?

zcorpan:
I think we should keep the id, expose it in the API and allow targeting it with Selectors. So I guess this PR is OK; I'll merge when fixing the related issues.

zcorpan:
I suspect you should use `git rebase` rather than `git merge`, but don't worry for this PR 🙂

silviapfeiffer:
Git and I ... Best friends, but sometimes... Sigh

silviapfeiffer:
Only relevant commit on this PR is 074059b7dc4eafcf6db2972857279b81c6d5d1aa

silviapfeiffer:
PR #349 replaces this, closing

PR #202, opened 2015-06-28
Expressly add the CSS boxes to the DOM.

silviapfeiffer:
Addresses https://www.w3.org/Bugs/Public/show_bug.cgi?id=23269

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -4265,7 +4265,8 @@ The Final Minute</pre>
      </li>
 
      <li><p>If there are no line boxes in <var>boxes</var>, skip the remainder of these substeps for
-     <var>cue</var>. The cue is ignored.</p></li>
+     <var>cue</var>. The cue is ignored. Otherwise, add <var>boxes</var> to the DOM to compute the
+     width and height of the line boxes.</p></li>
 
      <li>
 


foolip:
Unfortunately, *boxes* is a set of CSS boxes, because this is all written as if it were a layout algorithm operating on the layout tree, not the DOM tree. One cannot append CSS boxes to the DOM, but that is how to do it implementing in JavaScript using the DOM and not as part of layout.

In Blink, one can say that the algorithm up to "Adjust the positions of boxes according to the appropriate steps from the following list" is implemented using DOM manipulation while the adjustments themselves happen during layout. This isn't how the spec puts it, though.

I will comment on the bug.

silviapfeiffer:
So, reformulate as "add boxes to the layout tree"?

foolip:
That also wouldn't correspond to how one would actually implement this. One can add elements to the DOM which will on the next layout cause new layout boxes to be created, and there are things like the `::before` pseudo-element that aren't backed by an actual element in the DOM, but where the spec says "add boxes to the layout tree" one could implement it by anything like `layoutTree->add(boxes)`.

In order to solve this, it might be better if the spec expresses it more like how it's implemented. Namely, a DOM is constructed and inserted, and then during layout the positions of some of the boxes are adjusted to not overlap.

zcorpan:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=23269#c5

silviapfeiffer:
Fair enough :-)

PR #203, opened 2015-06-28
Clean up WebVTT timestamp spec.

silviapfeiffer:
Addresses https://www.w3.org/Bugs/Public/show_bug.cgi?id=25542

### DIFF
--- a/webvtt.html
+++ b/webvtt.html
@@ -1242,18 +1242,7 @@ Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% 
     cue block</a>. Different cues can overlap. Cues are always listed ordered by their start
     time.</p>
 
-    <p>A <dfn>WebVTT timestamp</dfn> can be either a <a title="complete WebVTT timestamp">WebVTT
-    timestamp representing hours, minutes, seconds and thousandths of a second</a> or a <a
-    title="partial WebVTT timestamp">WebVTT timestamp representing a time in seconds and fractions
-    of a second</a>.</p>
-
-    <p class="note">A <a>WebVTT timestamp</a> is always interpreted relative to the <a>current
-    playback position</a> of the media data that the WebVTT file is to be synchronized with, which
-    always starts at 0.</p>
-
-    <p>A <dfn title="complete WebVTT timestamp">WebVTT timestamp representing hours
-    <var>hours</var>, minutes <var>minutes</var>, seconds <var>seconds</var>, and thousandths of a
-    second <var>seconds-frac</var></dfn>, consists of the following components, in the given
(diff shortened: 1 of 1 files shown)


webvtt.html
```diff
@@ -1287,29 +1276,9 @@
...
-
-     <li><p>If <var>minutes</var> is greater than 59, then let <var>hours</var> be the integer
-     component of <var>minutes</var> divided by sixty, and then let <var>minutes</var> be the
-     remainder of dividing <var>minutes</var> divided by sixty. Otherwise, let <var>hours</var> be
-     zero.</p></li>
-
-    </ol>
+    <p class="note">A <a>WebVTT timestamp</a> is always interpreted relative to the <a>current
+    playback position</a> of the media data that the WebVTT file is to be synchronized with, which
+    always starts at 0.</p>
```
foolip: The "which always starts at 0" bit isn't guaranteed by the HTML spec I think:
https://html.spec.whatwg.org/#media-timeline

foolip: Oh, I see that this isn't actually new wording, just moved. I'll merge change fix separately.

dwsinger: This is a weird statement by the HTML spec.:

"In the rare case of a media resource that does not have an explicit timeline, “

No, the *common* case is that media is all relatively timed, relative to its start.  It’s *rare* that people embed e.g. SMPTE timecodes…

however, not our problem


> On Jul 20, 2015, at 7:05 , Philip Jägenstedt <notifications@github.com> wrote:
> 
> In webvtt.html:
> 
> > -     digits of the decimal fraction given to three decimal digits.</p></li>
> > -
> > -     <li><p>If <var>seconds</var> is greater than 59, then let <var>minutes</var> be the integer
> > -     component of <var>seconds</var> divided by sixty, and then let <var>seconds</var> be the
> > -     remainder of dividing <var>seconds</var> divided by sixty. Otherwise, let <var>minutes</var> be
> > -     zero.</p></li>
> > -
> > -     <li><p>If <var>minutes</var> is greater than 59, then let <var>hours</var> be the integer
> > -     component of <var>minutes</var> divided by sixty, and then let <var>minutes</var> be the
> > -     remainder of dividing <var>minutes</var> divided by sixty. Otherwise, let <var>hours</var> be
> > -     zero.</p></li>
> > -
> > -    </ol>
> > +    <p class="note">A <a>WebVTT timestamp</a> is always interpreted relative to the <a>current
> > +    playback position</a> of the media data that the WebVTT file is to be synchronized with, which
> > +    always starts at 0.</p>
> 
> The "which always starts at 0" bit isn't guaranteed by the HTML spec I think:
> https://html.spec.whatwg.org/#media-timeline
> 
> —
> 

David Singer

PR #206, opened 2015-08-09
Add a UTF8 character to a identifier...

silviapfeiffer:
...and show how it is referenced in CSS.
Closes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28256


zcorpan:
LGTM

PR #207, opened 2015-08-09
Add word change on Unicode parsing suggested by Addison.

silviapfeiffer:


zcorpan:
Should use lowercase "must not".

> For readability, these words do not appear in all uppercase letters in this specification.

PR #211, opened 2015-08-11
Remove "cue" from some definitions

foolip:
https://github.com/w3c/webvtt/pull/197#issuecomment-129745804


foolip:
Could we have this merged soon? fs is syncing with the spec in https://codereview.chromium.org/1306833003/ and it would be nice to only have to change it once.

silviapfeiffer:
Yup, sorry - busy few weeks.

foolip:
Thanks!

PR #212, opened 2015-09-22
Hand over to Simon

foolip:


foolip:
@zcorpan, your first job is to figure out how to get push access to this repo in order to merge this :)

https://github.com/foolip/webvtt-webhook is still running, but it sometimes needs some poking, so it would be nice if you or @silviapfeiffer took that over, or found another solution.

PR #213, opened 2015-09-24
Convert to bikeshed

zcorpan:
This generates without fatal errors but still with lots of warnings.


foolip:
There's some custom scripts left in webvtt.bs that have to do with ReSpec.

zcorpan:
@silviapfeiffer r?

silviapfeiffer:
How are you going to produce both the CG spec and the TTWG spec from this now that you've removed the TTWG spec text? Also, it's missing the styles of CG spec publication. Can we retain the CG spec styling with bikeshed? (I'm new to bikeshed, so just learning about it...)

zcorpan:
@silviapfeiffer when https://github.com/tabatkins/bikeshed/pull/498 is merged, the CG spec will have CG styling. A WD can be generated by changing the line `Status: CG-DRAFT` to `Status: WD` in the source. The WD will have the snapshot warning.

silviapfeiffer:
I'd like to land this basically once it creates http://dev.w3.org/html5/webvtt/ and http://www.w3.org/TR/webvtt1/ .

Things I'm still concerned about:
* the name of the relevant WG, which is different between the two specs
* the URLs to the relevant WGs
* wondering about the biblio refs...

zcorpan:
> I'd like to land this basically once it creates http://dev.w3.org/html5/webvtt/ and http://www.w3.org/TR/webvtt1/ .

Yeah I've tried to do that.

> the name of the relevant WG, which is different between the two specs

In "Status of This Document"? This should be the same as previously.

> the URLs to the relevant WGs

What do you mean?

> wondering about the biblio refs...

What about them?

zcorpan:
@silviapfeiffer ping?

silviapfeiffer:
>> I'd like to land this basically once it creates http://dev.w3.org/html5/webvtt/ and http://www.w3.org/TR/webvtt1/ .

> Yeah I've tried to do that.

OK, I'll see if I can replicate that today. It's the last thing I want to do before handing over.

> > the name of the relevant WG, which is different between the two specs

> In "Status of This Document"? This should be the same as previously.

OK, I'll check if the script does it all fine today.

> > the URLs to the relevant WGs

> What do you mean?

The groups out of which the WD and the draft come are different, so need different blurb.

> > wondering about the biblio refs...

> What about them?

The specific ones in the original config files - whether they got lost. Will also check today.

silviapfeiffer:
Actually looking pretty good.

A few things I noticed on the CG spec (in comparison to http://dev.w3.org/html5/webvtt/):
* the Copyright statement on the CG draft is very different - it needs to reference the CLA of the CG
* the "status of this document" section is missing
* when creating the WD, the images are not copied into the archive folder
* the CG-DRAFT needs to be created in the main directory and not (just) in archives, so it can be published through a gh-pages branch
* the reference to [MAUR] is lost

Additionally for the WD (in comparison to http://www.w3.org/TR/2014/WD-webvtt1-20141113/):
* the W3C layout seems to be cleaner (missing the sidebar of "W3C Working Draft" and the "W3C" logo at the top - I'm not sure if that's acceptable (I personally don't mind)
* the copyright is not the W3C copyright - again, I'm not sure if that's acceptable (I personally don't mind)
* the "Status of this document" is missing

Don't have the time right now to prepare a patch, but these are the things I found.

zcorpan:
How did you generate the CG spec? You should just run `bikeshed spec`, not use `snapshot.py`.

zcorpan:
Also have you updated bikeshed? The "Status of this document" and copyrights for both CG and WD should be there with up-to-date bikeshed.

zcorpan:
> * when creating the WD, the images are not copied into the archive folder
> * the reference to [MAUR] is lost
> * the W3C layout seems to be cleaner (missing the sidebar of "W3C Working Draft" and the "W3C" logo at the top - I'm not sure if that's acceptable (I personally don't mind)

Thanks, will fix these

silviapfeiffer:
> How did you generate the CG spec? You should just run `bikeshed spec`, not use `snapshot.sh`.

Yes, I used the `shapshot.sh`. Running `bikeshed spec` created only the CG spec and it overwrote the original webvtt.html file. How do I use the `bikeshed` command to create the two specs then? `bikeshed -h` wasn't very useful.

silviapfeiffer:
> Also have you updated bikeshed? The "Status of this document" and copyrights for both CG and WD should be there with up-to-date bikeshed.

Thanks - I did have to update it. I didn't know that those sections went into bikeshed itself! I seem to be able to create the CG spec now and it includes the right copyright and the SoD, but what about the WD?

zcorpan:
The idea is that you run ´bikeshed spec` to generate the CG spec, and run `snapshot.sh` to generate a snapshot. I can clarify this in the README.

silviapfeiffer:
LGTM. Let's create that gh-pages branch!

PR #214, opened 2015-09-30
[issue 29165] updated to clarify positioning relationship to ruby.

ckennedy44:
closes issue 29165: https://www.w3.org/Bugs/Public/show_bug.cgi?id=29165


silviapfeiffer:
WFM.

PR #215, opened 2015-10-01
proposed fix for Bug 28267- remove work-wrap leaving only overflow-wrap

ckennedy44:
Based on the comments in https://www.w3.org/Bugs/Public/show_bug.cgi?id=28267, I think that this issue can be resolved with the patch included here.  

In the section of the document that describes the CSS settings that every WebVTT region is initialized with, the word-wrap property line can be removed . The overflow-wrap line is already included.


zcorpan:
Thanks, comitted as 43dd2f70eeb95ec7d9e551337ad81a5e0d7f0479

PR #219, opened 2015-10-19
Add support for STYLE blocks (inline style)

zcorpan:
This is Simon Pieter's proposal with only minor typo fixes:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=15023#c31

------

This is @foolip's PR #185, only bikeshedded (prose is the same).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -19,7 +19,7 @@ Abstract: This specification defines WebVTT, the Web Video Text Tracks format. I
 Abstract: WebVTT files provide captions or subtitles for video content, and also text video descriptions [[MAUR]], chapters for content navigation, and more generally any form of metadata that is time-aligned with audio or video content.
 Boilerplate: omit conformance, omit feedback-header
 Ignored Terms: unicode-bidi, color, text-combine-upright
-Ignored Vars: seconds-frac, selector, fragment
+Ignored Vars: seconds-frac, selector, fragment, seen cue
 
 </pre>
 
@@ -285,6 +285,88 @@ one line except when a line break is definitely necessary.</p>
  </pre>
 
 </div>
+
+<h3 id=styling>Styling</h3>
+
+<p><i>This section is non-normative.</i></p>
+
+<p>CSS stylesheets that apply to an HTML page that contains a <a spec=html element>video</a> element
+can target WebVTT cues and regions in the video using the ''::cue'', ''::cue()'' and
+''::cue-region'' pseudo-elements.</p>
+
+<div class="example">
+
+ <p>In this example, an HTML page has a CSS stylesheet in a <a spec=html element>style</a> element
+ that styles all cues in the video with a gradient background and a text color, as well as changing
+ the text color for all <a>WebVTT Bold Objects</a> in cues in the video.</p>
+
+ <pre>
+ &lt;!doctype html>
+ &lt;html>
+  &lt;head>
+   &lt;title>Styling WebVTT cues&lt;/title>
+   &lt;style>
+    video::cue {
+      background-image: linear-gradient(to bottom, dimgray, lightgray);
+      color: papayawhip;
+    }
+    video::cue(b) {
+      color: peachpuff;
+    }
+   &lt;/style>
+  &lt;/head>
+  &lt;body>
+   &lt;video controls autoplay src="video.webm">
+    &lt;track default src="track.vtt">
+   &lt;/video>
+  &lt;/body>
+ &lt;/html>
+ </pre>
+
+</div>
+
+<p>CSS stylesheets can also be embedded in WebVTT files themselves.</p>
+
+<p>Style blocks are placed after any headers but before the first cue, and start with the line
+"STYLE". Comment blocks can be interleaved with style blocks.</p>
+
+<p>Blank lines cannot appear in the stylesheet. They can be removed or be filled with a space or a
+CSS comment (e.g. <code>/**/</code>).</p>
+
+<p>The string "<code>--></code>" cannot be used in the stylesheet. If the stylesheet is wrapped in
+"<code>&lt;!--</code>" and "<code>--></code>", then those strings can just be removed. If
+"<code>--></code>" appears inside a CSS string, then it can use CSS escaping e.g.
+"<code>--\></code>".</p>
+
+<div class="example">
+
+ <p>This example shows how cues can be styled with style blocks in WebVTT.</p>
+
+ <pre>
+ WEBVTT
+
+ STYLE
+ ::cue {
+   background-image: linear-gradient(to bottom, dimgray, lightgray);
+   color: papayawhip;
+ }
+ /* Style blocks cannot use blank lines nor "dash dash greater than" */
+
+ NOTE comment blocks can be used between style blocks.
+
+ STYLE
+ ::cue(b) {
+   color: peachpuff;
+ }
+
+ hello
+ 00:00:00.000 --> 00:00:10.000
+ Hello &lt;b>world&lt;/b>.
+
+ NOTE style blocks cannot appear after the first cue.
+ </pre>
+
+ </div>
 
 <h3 id=introduction-other-features>Other features</h3>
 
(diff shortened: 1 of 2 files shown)


zcorpan:
- [ ] Tweak introduction to match the style of the other sections
- [ ] Fix base URL <blockquote>
(In reply to Simon Pieters from comment #31)
<blockquote> For the purpose of resolving URLs in the style sheet, the style sheet must
 act as if its base URL is the same as the WebVTT file that contains it. (For
 out-of-band text tracks, this is the _track URL_. For in-band text tracks,
 this is the URL of the containing _media resource_. Text tracks created with
 addTextTrack() cannot have style sheets.)</blockquote>
Actually _track URL_ is before redirects. I suppose we want the base URL to be after any redirects.
</blockquote>
- [ ] Remove `type` setting <blockquote>require that the line matches /^STYLE\s*$/, so that if we introduce settings in the future, legacy clients will drop the block.</blockquote>
- [ ] Fix CSSOM/css-syntax integration<blockquote>
The integration with CSS parsing seems wrong: http://dev.w3.org/csswg/css-syntax/#parse-a-stylesheet returns a stylesheet, not rules to attach to an existing stylesheet. Did I find the wrong thing?
</blockquote>
- [ ] Fix missing reference "valid css stylesheet"
<blockquote>I think we can say that the string represents a CSS stylesheet, and so the requirements in the CSS specs apply.</blockquote>
- [x] Processing model is difficult to implement<blockquote>The processing model seems difficult to implement. All cues from all tracks end up in the same container (since they must avoid overlapping each other) and it would be nice if this could internally be implemented by inserting a &lt;style> element with the relevant style into the shadow DOM that contains all the cues. I don't know how to isolate the style to individual tracks, maybe one container per track but that makes the overlap avoidance slightly uglier.</blockquote>
AIUI from the bug discussion, the current spec text is desired. May revisit when we have impl experience.
- [ ] Use "featureless"<blockquote>
(In reply to Simon Pieters from comment #31)
<blockquote> For the purpose of selectors in STYLE blocks of a WebVTT file, the style
 sheet must apply to a hypothetical document that contains a single empty
 element with no explicit name, no namespace, no attributes, no classes, no
 IDs, and unknown primary language, that acts like the _media element_ for
 the _text tracks_ that were sourced from the given WebVTT file. The
 selectors must not match other _text tracks_ for the same _media element_.</blockquote>
This element can now be described as "featureless" that can only be matched when using ::cue, ::cue() and ::cue-region.
http://dev.w3.org/csswg/selectors-4/#featureless</blockquote>

index.bs
```diff
@@ -1850,6 +1942,103 @@ stream lacks this WebVTT file signature, then the parser aborts.</p>
  <li><p>If <var>line</var> is not the empty string, then jump back to the step labeled
  <i>header</i>.</p></li>
 
+ <li><p><i>Style block loop</i>: <a spec=html>Collect a sequence of characters</a> that are U+000A
```
foolip: Rather than a style block loop followed by a cue loop, I think we should have a single loop that handles style blocks and cues interleaves, but discards style blocks that come after any cue. This will make it trivially easy to allow interleaved blocks when we want them, and gives us high confidence that we won't stumble upon some mistake in the structure of the parser only then, when it may be too late to change.

dwsinger: > On Oct 22, 2015, at 9:42 , Philip Jägenstedt <notifications@github.com> wrote:
> 
> In index.bs:
> 
> > @@ -1850,6 +1942,103 @@ stream lacks this WebVTT file signature, then the parser aborts.</p>
> >   <li><p>If <var>line</var> is not the empty string, then jump back to the step labeled
> >   <i>header</i>.</p></li>
> >  
> > + <li><p><i>Style block loop</i>: <a spec=html>Collect a sequence of characters</a> that are U+000A
> 
> Rather than a style block loop followed by a cue loop, I think we should have a single loop that handles style blocks and cues interleaves, but discards style blocks that come after any cue. This will make it trivially easy to allow interleaved blocks when we want them, and gives us high confidence that we won't stumble upon some mistake in the structure of the parser only then, when it may be too late to change.
> 

sounds good and prudent


David Singer

index.bs
```diff
@@ -270,6 +270,76 @@ one line except when a line break is definitely necessary.</p>
...
+
+<div class="example">
+
+ <p>This example shows how cues can be styled with style blocks in WebVTT.</p>
+
+ <pre>
+ WEBVTT
+
+ STYLE
+ ::cue(#hello) { color:lime }
```
silviapfeiffer: Might be worth simply using the same styling example in the HTML page and the embedded example to show equivalence. Not a biggie, just a though.

index.bs
```diff
@@ -270,6 +270,76 @@ one line except when a line break is definitely necessary.</p>
...
+ /* Style blocks cannot use blank lines nor "dash dash greater than" */
+
+ NOTE comment blocks can be used between style blocks.
+
+ STYLE
+ ::cue { font-variant:small-caps }
+
+ hello
+ 00:00:00.000 --&gt; 00:00:10.000
```
silviapfeiffer: You can use the ">" instead of the escaped character here - is more consistent with the other examples.

index.bs
```diff
@@ -270,6 +270,76 @@ one line except when a line break is definitely necessary.</p>
...
+  &lt;/body>
+ &lt;/html>
+ </pre>
+
+</div>
+
+<p>CSS stylesheets can also be embedded in WebVTT files themselves.</p>
+
+<p>Style blocks are placed after any headers but before the first cue, and start with the line
+"STYLE". Blank lines and the string "<code>--></code>" cannot appear in the stylesheet. Comment
```
silviapfeiffer: Might be worth explaining to use "--&gt;" instead of "-->" inside a style in the STYLE block, since "-->" ends the STYLE block.

silviapfeiffer:
Great work. No biggies from me, just some recommendations of improving wording.

index.bs
zcorpan: https://github.com/w3c/webvtt/commit/e02c236e0914a563b026f896a404eddadf8d407f

index.bs
zcorpan: https://github.com/w3c/webvtt/commit/efb3385847f684bab688b93bd50a6dddfa556915

index.bs
zcorpan: It doesn't discard the block, it just ends the block and starts a new cue. But yeah, https://github.com/w3c/webvtt/commit/adcdbf12e20413e99e908aa209b2e40030c03e78

silviapfeiffer:
Are you going to address Phillip's comment in this patch or a different one?

zcorpan:
Philip's comment is https://github.com/w3c/webvtt/issues/222 -- then plan is to rebase this on top of that.

zcorpan:
A thing I noticed while looking at rewriting the parser is the following case:

```
WEBVTT

STYLE
00:00:00.000 --> 00:00:01.000
test
```

Currently this is a perfectly valid cue with the id "STYLE". It should continue to be so. (Similarly with NOTE.)

zcorpan:
Need to define what the relative order of stylesheets should be for the cascade, also when using both in-file CSS and styles in the parent HTML doc.

zcorpan:
Depends on: #246

zcorpan:
OK squashed and rebased on top of the new parser, which should address @foolip's comment.

zcorpan:
I think the featureless thing should be reverted, because when an element is featureless but can be selected with a peudo-element selector, it implies that the element we're selecting is the pseudo-element itself. But that's not what we want here, we want the originating element of `::cue` to not be generally selectable. cc @tabatkins

https://drafts.csswg.org/selectors-4/#pseudo-element%20syntax

> Pseudo-elements are featureless, and so can’t be matched by any other selector.

zcorpan:
All comments addressed, please review!

- [HTML output](http://rawgit.com/w3c/webvtt/inline-style/index.html)
- [HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fw3c%2Fwebvtt%2Fcfef0bce5dde5bd4efeb4444d6187ff5a39f0827%2Findex.html&doc2=http%3A%2F%2Frawgit.com%2Fw3c%2Fwebvtt%2Finline-style%2Findex.html)

zcorpan:
Is someone willing to review this anytime soon? If not, I suggest we merge and fix any issues later.

silviapfeiffer:
Had a quick look through and looks ok to me.

zcorpan:
Thanks @silviapfeiffer

silviapfeiffer:
I was kinda  hoping somebody else would chip in because I had already given my ok, but I'm glad it's done and this is a big step ahead! So good to have you take charge.

PR #223, opened 2015-10-24
Refer to the HTML spec for which IDL interface elements use

zcorpan:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=22154.


silviapfeiffer:
LGTM

PR #237, opened 2015-11-05
Fix #232: Make "text alignment" definitions match the rendering

### LINKED ISSUE #232: Definition of "text alignment" doesn't match processing requirements
https://w3c.github.io/webvtt/#webvtt-cue-text-alignment

> A text alignment
An alignment for all lines of text within the cue box, in the dimension of the writing direction and the paragraph direction [BIDI], one of:

> Start alignment
The text is aligned towards the paragraph direction start side of the cue box.

As far as I can tell, this text is wrong. The [processing model](https://w3c.github.io/webvtt/#applying-css-properties) sets 'unicode-bidi' to `plaintext` which isolates paragraphs, and then sets 'text-align' to `start` for "Start alignment", and the behavior of CSS in that case does not match the description above.

See
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3719
and flip the value of 'text-align' between `start` and `end` (in Gecko or WebKit or Blink), to see what the effect of the processing requirements are.

So for start alignment, each line is aligned to the start edge for that line, which is separate from the **cue**'s "paragraph direction".

My understanding is that @r12a is happy with the effect of the processing requirements, so I suggest we fix the description here to match.

zcorpan:
The rendering rules apply unicode-bidi:plaintext which isolates
each line. Make the definitions for "text alignment" start/end
match the rendering rules. Drop the "paragraph direction" concept
since it was not used for anything that changes the rendered result.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -18,7 +18,7 @@ Test Suite: https://github.com/w3c/web-platform-tests/tree/master/webvtt
 Abstract: This specification defines WebVTT, the Web Video Text Tracks format. Its main use is for marking up external text track resources in connection with the HTML &lt;track> element.
 Abstract: WebVTT files provide captions or subtitles for video content, and also text video descriptions [[MAUR]], chapters for content navigation, and more generally any form of metadata that is time-aligned with audio or video content.
 Boilerplate: omit conformance, omit feedback-header
-Ignored Terms: unicode-bidi, direction, color, text-combine-upright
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
WFM. Also: loving your activity! Great work and great progress!

zcorpan:
Thanks!

PR #238, opened 2015-11-05
Fix #233: Parse a percentage string should fail for invalid input

### LINKED ISSUE #233: "parse a percentage string" parses ".1%" and "1.%"
https://w3c.github.io/webvtt/#parse-a-percentage-string

> parse a percentage string

This algorithm doesn't fail for the strings  ".1%" and "1.%". Is that intentional? I think it should fail if it doesn't match https://w3c.github.io/webvtt/#webvtt-percentage

zcorpan:
The algorithm previously parsed ".1%" and "1.%" which is not consistent
with either HTML or CSS.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2115,19 +2115,7 @@ means that it is aborted at that point and returns nothing.</p>
 <ol>
  <li><p>Let |input| be the string being parsed.</p></li>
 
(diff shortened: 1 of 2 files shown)


foolip:
:+1:

PR #240, opened 2015-11-05
Fix #236: Make <rt> be bidi-isolated

### LINKED ISSUE #236: Isolate bidi for ruby text?
Should the text in `<rb>` be isolated for the purposes of bidi? It seems to me that it should be possible to use e.g. English ruby text for Arabic ruby base.

    <ruby>مرحبا!<rt>Hello!</ruby>

I think setting 'unicode-bidi' to "plaintext" for `<rt>` would fix this.

CC @r12a

zcorpan:
This makes it possible to mix RTL and LTR between ruby base and
ruby text and have it just work. `&lrm;` and `&rlm;` in `<rt>` will also
work to override if necessary.


silviapfeiffer:
FWIW: WFM :-)

PR #243, opened 2015-11-05
Add 'text-combine-upright' to the list of supported CSS properties

zcorpan:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28184.


zcorpan:
cc @dae-kim

silviapfeiffer:
WFM, FWIW

PR #244, opened 2015-11-05
Rename "middle" to "center" throughout for consistency with CSS

zcorpan:
This affects align, position align, line align, and the DOM API.

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=25956.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -466,7 +466,7 @@ CSS comment (e.g. <code>/**/</code>).</p>
  00:00:03.000 --> 00:00:06.500 position:90% align:right size:35%
  I think he went down this lane.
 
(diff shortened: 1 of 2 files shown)


zcorpan:
cc @eric-carlson @rickeyre @foolip

foolip:
Since I filed the bug, it should come as not surprise that I support this. I've skimmed the diff and assume it's just search+replace, so :+1:

zcorpan:
It's search+replace but I sanity-checked each instance in doing so.

I'd like a :+1: from @eric-carlson or @rickeyre (or other active implementor) before merging this.

eric-carlson:
:+1: from me. Thanks Simon!

zcorpan:
Great, thanks :-)

PR #246, opened 2015-11-06
Fix #222: Revamp the parser

### LINKED ISSUE #222: Revamp the parser
To fix https://github.com/w3c/webvtt/pull/219#discussion_r42719065 I find the flat structure of the parsing algorithm with "jump to step foo" makes things hard to work with. Explore a different way to specify the parser that can e.g. share steps for skipping to the next blank line or "-->", and makes it easier to add new kinds of blocks.

This should be purely editorial.

Also see http://krijnhoetmer.nl/irc-logs/whatwg/20151022#l-301 for some discussion.

### LINKED ISSUE #224: timings in header followed by a blank line is not parsed correctly
https://w3c.github.io/webvtt/#file-parsing

> If the character indicated by position is a U+000A LINE FEED (LF) character, advance position to the next character in input.

Consider this file:

```
WEBVTT
00:00:00.000 --> 00:00:01.000

Test
```

This parses into a cue with cue text "Test" per spec, AFAICT. This looks like a bug.

Metadata header loop
> 14 The character indicated by position is a U+000A LINE FEED (LF) character. Advance position to the next character in input.

> 15 If line contains the three-character substring "-->" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN), then set the already collected line flag and jump to the step labeled cue loop.

Now position is at the blank line.

> 17 Cue loop: If the already collected line flag is set, then jump to the step labeled cue creation.

Cue creation:

> 22 If line contains the three-character substring "-->" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN), then jump to the step labeled timings below.

> 28 Timings: Unset the already collected line flag.

> 29 Collect WebVTT cue timings and settings from line using regions for cue. If that fails, jump to the step labeled bad cue.

didn't fail...

> 31 Cue text loop: If position is past the end of input, then jump to the step labeled cue text processing.

> 32 If the character indicated by position is a U+000A LINE FEED (LF) character, advance position to the next character in input.

This is the bug. This skips past the blank line so position is at the start of the "Test" line.

zcorpan:
Introduce a new "consume a WebVTT block" concept that is used
for parsing the header, cues, and discarding of bad cues.

This should be strictly editorial except that it also fixes #224.
It should be easier to add new block types like STYLE, and hopefully
easier to implement without having the algorithm use GOTO everywhere.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1887,7 +1887,7 @@ stream lacks this WebVTT file signature, then the parser aborts.</p>
  to |input|, then the |position| pointer may, when so instructed by the algorithms, be moved past
  the end of |input|.</p></li>
 
- <li>Let |line| be a string variable. Unset the |already collected line| flag.</li>
+ <li>Let |line| be a string variable.</li>
 
  <!-- SIGNATURE CHECK -->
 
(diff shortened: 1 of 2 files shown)


zcorpan:
cc @RickEyre @andreasgal @humphd @foolip @eric-carlson

(Also see the issue: #222)

zcorpan:
- [HTML output](https://rawgit.com/w3c/webvtt/revamp-parser/index.html)
- [HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fw3c%2Fwebvtt%2Fbda95b7e5c536c943382a8b5d50be1aac4dd0107%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fwebvtt%2Frevamp-parser%2Findex.html)

zcorpan:
Also see #254 which would make the parser less awkward.

zcorpan:
Status: fs has done a high-level review and thought the approach was OK, but would have to actually implement it to verify completely, but no plan to do so at this time. I don't see anyone else stepping forward to review this at this time either.

Since I have written an implementation myself and compared with the old algorithm, I am relatively confident about the correctness, so I will go ahead and merge this to unblock the other issues.

PR #247, opened 2015-11-09
Avoid waiting for a LF before checking the WEBVTT signature

zcorpan:
This allows the parser to fail fast for e.g. a long stream of data
that doesn't contain LF.

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=22315.


silviapfeiffer:
WFM

PR #248, opened 2015-11-11
Fix #227: Explain how bidi works in WebVTT

### LINKED ISSUE #227: Explain the difference between cue direction and paragraph direction
See https://www.w3.org/Bugs/Public/show_bug.cgi?id=28266

The processing model is OK but it is very unclear what the effect is. We should add a note clarifying this, with an example. In particular clarify that the cue direction (which is determined by the direction of the first paragraph in the cue) affects positioning of the cue, but unicode-bidi:plaintext still makes the direction of each paragraph be isolated from each other.

Also see http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3718

zcorpan:
Add lots of notes and examples related to bidi issues. Also change
the style of sample cue rendering to make it stand out more.

The text is partially based on Richard Ishida's text in
https://github.com/w3c/webvtt/issues/227#issuecomment-154486613.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -91,6 +91,16 @@ spec:css-flexbox-1; type:value; text:inline-flex
 }
 </pre>
 
+<style>
+ samp {
+   font-family: inherit;
+   background-color: black; /* fallback if rgba() is not supported */
+   background-color: rgba(0, 0, 0, 0.7);
+   outline: 0.18em solid rgba(0, 0, 0, 0.7);
+   color: white;
+ }
+</style>
+
 <h2 id=introduction>Introduction</h2>
 
 <p><i>This section is non-normative.</i></p>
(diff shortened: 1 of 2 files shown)


zcorpan:
- [HTML output](https://rawgit.com/w3c/webvtt/clarify-direction/index.html)
- [HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Frawgit.com%2Fw3c%2Fwebvtt%2F7ebfa9ad31aba0c261c067acf33775d508b47af3%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fwebvtt%2Fclarify-direction%2Findex.html)

silviapfeiffer:
LGTM - I like the examples inline, though mostly we've been doing them at the beginning this far.

PR #253, opened 2015-11-12
Support all of HTML's character references

zcorpan:
This hooks in to HTML's 'consume a character reference' for parsing,
and 'character references' for the syntax. All forms of charrefs
that HTML supports are supported, including missing semicolon.
(But missing semicolon is not valid, nor is a lone ampersand.)

Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=23867.
Ref. #198.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -53,6 +53,8 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: text track cue end time
             text: expose a user interface to the user
             text: text track cue order
+        urlPrefix: syntax.html
+            text: syntax-charref
     type: element-attr
         urlPrefix: dom.html
             text: title; url: #attr-title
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
WFM

zcorpan:
@annevk can I ask for a sanity check from HTML's point of view?

annevk:
I would have preferred a truly reusable algorithm rather than this monkey patch way, but I guess it's okay.

zcorpan:
OK, we can clean it up when we have implementation experience suggesting this approach is working. Thanks!

zcorpan:
https://bugs.chromium.org/p/chromium/issues/detail?id=761303
https://issuetracker.google.com/issues/65278653
https://bugs.webkit.org/show_bug.cgi?id=176225
https://bugzilla.mozilla.org/show_bug.cgi?id=1395924
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13551960/

PR #256, opened 2015-11-13
Fix #217: Clarify that only validators will validate language tags

### LINKED ISSUE #217: Clarify "Valid BCP47 language tag"
> In http://dev.w3.org/html5/webvtt/#webvtt-cue-text you still say the cue language span "must be a valid BCP 47 language tag", where 'valid' has a particular meaning in BCP 47 and it is not clear if you intend that specific meaning. You should: (a) omit the word valid; (b) clarify that you mean BCP 47 valid; or (c) use the word 'well-formed' instead. (Valid in BCP 47 means that the implementation checks to see that all of the subtags are registered, at least as of an implementation specific date)

I think (b) is intended. HTML has the same requirement for its `lang` attribute. But it's not a requirement on implementations in HTML and WebVTT, but on authors.

https://www.w3.org/Bugs/Public/show_bug.cgi?id=28255#c17

zcorpan:
See #217

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1472,6 +1472,10 @@ given:</p>
  <li>A <a>WebVTT cue span end tag</a> "<code>lang</code>".</li>
 </ol>
 
+<p class=note>The requirement above regarding valid BCP 47 language tag is an authoring requirement,
+so a conformance checker will do validity checking of the language tag, but other user agents will
+not.</p>
+
 <p>A <dfn>WebVTT cue span start tag</dfn> has a |tag name| and either <!--allows,-->
 requires<!--,--> or disallows an annotation, and consists of the following components, in the order
 given:</p>
@@ -2841,6 +2845,10 @@ Object">WebVTT Internal Node Objects</a> also have an ordered list of class name
 <dfn lt="WebVTT Node Object's applicable classes">applicable classes</dfn>, and a language, known as
 their <dfn lt="WebVTT Node Object's applicable language">applicable language</dfn>, which is to be
 interpreted as a BCP 47 language tag. [[!BCP47]]</p>
+
+<p class=note>User agents will add a language tag as the <a lt="WebVTT Node Object's applicable
+language">applicable language</a> even if it is not a valid or not even well-formed language tag.
+[[!BCP47]]</p>
 
 <p>There are several concrete classes of <a lt="WebVTT Internal Node Object">WebVTT Internal Node
 Objects</a>:</p>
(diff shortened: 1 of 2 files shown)


zcorpan:
cc @aphillips

zcorpan:
Ping @aphillips @r12a

index.bs
```diff
@@ -2842,6 +2846,10 @@ Object">WebVTT Internal Node Objects</a> also have an ordered list of class name
 their <dfn lt="WebVTT Node Object's applicable language">applicable language</dfn>, which is to be
 interpreted as a BCP 47 language tag. [[!BCP47]]</p>
 
+<p class=note>User agents will add a language tag as the <a lt="WebVTT Node Object's applicable
```
aphillips: I understand the goal of this note, but it seems like it isn't worded correctly, since we want to encourage user agents to supply a well-formed and valid language tag, but not require them to check the language tag. I would probably say:

```
User agents must add a language tag as the applicable language. While language tag should be a valid language tag [[!BCP47]], the user agent is not required to check the structure or contents of the language tag to ensure that this is so.
```

index.html
```diff
@@ -3160,6 +3163,7 @@ <h3 class="heading settled" data-algorithm="WebVTT cue text parsing rules" data-
 parent-child relationships so constructed form a tree structure. <a data-link-type="dfn" href="#webvtt-internal-node-object">WebVTT Internal Node Objects</a> also have an ordered list of class names, known as their <dfn data-dfn-type="dfn" data-lt="WebVTT Node Object’s applicable classes" data-noexport="" id="webvtt-node-objects-applicable-classes">applicable classes<a class="self-link" href="#webvtt-node-objects-applicable-classes"></a></dfn>, and a language, known as
 their <dfn data-dfn-type="dfn" data-lt="WebVTT Node Object’s applicable language" data-noexport="" id="webvtt-node-objects-applicable-language">applicable language<a class="self-link" href="#webvtt-node-objects-applicable-language"></a></dfn>, which is to be
 interpreted as a BCP 47 language tag. <a data-link-type="biblio" href="#biblio-bcp47">[BCP47]</a></p>
+   <p class="note" role="note">User agents will add a language tag as the <a data-link-type="dfn" href="#webvtt-node-objects-applicable-language">applicable language</a> even if it is not a valid or not even well-formed language tag. <a data-link-type="biblio" href="#biblio-bcp47">[BCP47]</a></p>
```
aphillips: See above.

index.bs
zcorpan: Except notes are non-normative and shouldn't use normative keywords (should, required).

PR #257, opened 2015-11-13
Allow external language information to apply

zcorpan:
Make it possible for `<track srclang>` (or other external language)
to apply to the WebVTT nodes for the purpose of CSS `:lang()` matching,
glyph selection, etc.

"WebVTT cue text parsing rules" takes an optional fallback language.

The "rules for updating the display of WebVTT text tracks" also takes
an optional fallback language, since it is the entry point for HTML
and this algorithm then invokes the "WebVTT cue text parsing rules".

`getCueAsHTML()` does not provide a fallback language.

Fixes (1) in https://www.w3.org/Bugs/Public/show_bug.cgi?id=28255#c16.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2621,7 +2621,7 @@ user agent must run the following steps:</p>
 </ol>
 
 
-<h3 id=cue-text-parsing-rules algorithm><dfn>WebVTT cue text parsing rules</dfn></h3>
+<h3 id=cue-text-parsing-rules algorithm>WebVTT cue text parsing rules</h3>
 
 <p>A <dfn>WebVTT Node Object</dfn> is a conceptual construct used to represent components of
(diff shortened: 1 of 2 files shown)


zcorpan:
Should the "List of WebVTT Node Objects" object have a "lang" attribute for the purpose of Selectors? (Similar to `<lang ...>`.)

https://w3c.github.io/webvtt/#selectordef-cue-selector

> For the purposes of attribute selector matching, ... WebVTT Language Objects, which have a single attribute named "lang" whose value is the object’s applicable language.

silviapfeiffer:
LGTM. To hand down the `lang` setting to the suggested `DocumentFragment`'s root node, you indeed need to keep it in the "List of WebVTT Node Objects" object.

PR #263, opened 2015-11-16
Add 'ruby-position' to the lists of supported properties

zcorpan:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28183.


silviapfeiffer:
That was easy :-) LGTM

zcorpan:
I added it to both `::cue`'s and `::cue()`'s (sans-`:past`/`:future`) lists since the property is inherited, it can make sense to set it on the root.

PR #266, opened 2015-11-16
Fix #252: Actually parse charrefs in annotation

### LINKED ISSUE #252: Tokenizer doesn't parse escapes in annotation
https://w3c.github.io/webvtt/#webvtt-start-tag-annotation-state

> WebVTT start tag annotation state
> Jump to the entry that matches the value of c:
>
> U+003E GREATER-THAN SIGN character (>)
> Advance position to the next character in input, then jump to the next "end-of-file marker" entry below.
>
> End-of-file marker
> Remove any leading or trailing space characters from buffer, and replace any sequence of one or more consecutive space characters in buffer with a single U+0020 SPACE character; then, return a start tag whose tag name is result, with the classes given in classes, and with buffer as the annotation, and abort these steps.
>
> Anything else
> Append c to buffer and jump to the step labeled next.

This doesn't tokenize escapes, AFAICT. But the syntax allows escapes in annotations:

https://w3c.github.io/webvtt/#webvtt-cue-span-start-tag

>If the start tag requires an annotation: a U+0020 SPACE character or a U+0009 CHARACTER TABULATION (tab) character, followed by one or more of the following components, the concatenation of their representations having a value that contains at least one character other than U+0020 SPACE and U+0009 CHARACTER TABULATION (tab) characters:
>
> WebVTT cue span start tag annotation text, representing the text of the annotation.
> A WebVTT cue amp escape, representing a "&" character in the text of the annotation.
> A WebVTT cue lt escape, representing a "<" character in the text of the annotation.
> A WebVTT cue gt escape, representing a ">" character in the text of the annotation.
> A WebVTT cue lrm escape, representing a U+200E LEFT-TO-RIGHT MARK Unicode bidirectional formatting character in the text of the cue.
> A WebVTT cue rlm escape, representing a U+200F RIGHT-TO-LEFT MARK Unicode bidirectional formatting character in the text of the cue.
> A WebVTT cue nbsp escape, representing a U+00A0 NO-BREAK SPACE character in the text of the cue.

zcorpan:
The syntax allowed charrefs in annotations, but the tokenizer didn't.

Rename "HTML character reference state" to "HTML character reference
in data state" and add a new "HTML character reference in annotation
state". Switch into the new state when seeing "&" in the "WebVTT
start tag annotation state".

#252

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3159,8 +3159,8 @@ value).</p>
 
      <dt>U+0026 AMPERSAND (&amp;)</dt>
      <dd>
-      <p>Set |tokenizer state| to the <a>HTML character reference state</a>, and jump to
-      the step labeled <i>next</i>.</p>
+      <p>Set |tokenizer state| to the <a>HTML character reference in data state</a>, and jump to the
+      step labeled <i>next</i>.</p>
      </dd>
 
      <dt>U+003C LESS-THAN SIGN (&lt;)</dt>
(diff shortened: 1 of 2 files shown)


zcorpan:
@foolip can you review?

index.bs
```diff
@@ -3190,6 +3190,22 @@ value).</p>
 
    </dd>
 
+   <dt><dfn>HTML character reference in annotation state</dfn></dt>
+
+   <dd>
+
+    <p>Attempt to <a>consume an HTML character reference</a>, with the <a spec=html>additional
```
foolip: I think this is right, because HTML says "optionally with an additional allowed character, which, if specified where the algorithm is invoked, adds a character to the list of characters that cause there to not be a character reference."

Without any context, I assumed this meant that `>` would be allowed as part of the character reference, which would the opposite of the intended.

foolip:
:+1:

index.bs
zcorpan: Yeah the only difference from "no additional allowed character" is whether `&>` produces a parse error or not. WebVTT doesn't take note of the parse error, so for WebVTT there is technically no difference at all, currently. WebVTT's syntax doesn't allow lone ampersands, so it is a syntax violation regardless.

zcorpan: Seems there's no actual difference... whatwg/html#364

PR #267, opened 2015-11-17
Fix #228: Add an example with classes and language change

### LINKED ISSUE #228: Clarify that <b> can use a class name and CSS
https://w3c.github.io/webvtt/#webvtt-bold-object

> WebVTT Bold Objects

See https://www.w3.org/Bugs/Public/show_bug.cgi?id=28262

Also http://www.w3.org/International/questions/qa-b-and-i-tags (the Japanese example)

For HTML @r12a said that the way to handle this example in HTML is to use span or b with a class name, and use CSS to style it differently for different languages. So we should clarify that this is possible in WebVTT as well and add an example of how to do that for this in particular.

zcorpan:
#228 

cc @r12a

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -402,6 +402,23 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 <div class="example">
 
+ <p>This example shows how classes can be used on elements, which can be helpful for localization or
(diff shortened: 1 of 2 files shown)


zcorpan:
@r12a do you want to review this?

r12a:
:+1:  thanks

PR #273, opened 2015-11-27
Make position alignment saner

zcorpan:
* Make align:start/end not affect position or position alignment.
* Rename start/end for position alignment to left/right.
* Clarify what align:left/right do for vertical text.

Fixes:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28257#c9
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28257#c11
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28257#c13

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -461,13 +461,13 @@ CSS comment (e.g. <code>/**/</code>).</p>
  <pre>
  WEBVTT
 
- 00:00:00.000 --> 00:00:04.000 position:10%,start align:left size:35%
+ 00:00:00.000 --> 00:00:04.000 position:10%,line-left align:left size:35%
  Where did he go?
 
  00:00:03.000 --> 00:00:06.500 position:90% align:right size:35%
  I think he went down this lane.
 
- 00:00:04.000 --> 00:00:06.500 position:45%,end align:center size:35%
+ 00:00:04.000 --> 00:00:06.500 position:45%,line-right align:center size:35%
  What are you waiting for?
  </pre>
 
@@ -475,31 +475,34 @@ CSS comment (e.g. <code>/**/</code>).</p>
  of the width of the video viewpoint. If the text were vertical, the "position" setting would refer
  to the height of the viewport.</p>
 
- <p>The "start" or "end" only refers to the physical side of the box to which the "position" setting
- applies, in a way which is agnostic regarding the horizontal or vertical direction of the cue. It
- does not affect or relate to the direction or position of the text itself within the box.</p>
+ <p>The "line-left" or "line-right" only refers to the physical side of the box to which the
+ "position" setting applies, in a way which is agnostic regarding the horizontal or vertical
+ direction of the cue. It does not affect or relate to the direction or position of the text itself
+ within the box.</p>
 
  <p>The cues cover only 35% of the video viewport's width - that's the <a lt="WebVTT cue box">cue
  box</a>'s "size" for all three cues.</p>
 
- <p>The first cue has its <a lt="WebVTT cue box">cue box</a> positioned at the 10% mark. The "start"
- and "end" within the "position" setting indicates which side of the <a lt="WebVTT cue box">cue
- box</a> the position refers to. Since in this case the text is horizontal, "start" refers to the
- left side of the box, and the cue box is thus positioned between the 10% and the 45% mark of the
- video viewport's width, probably underneath a speaker on the left of the video image. If the cue
- was vertical, "start" positioning would be from the top of the video viewport's height and the <a
- lt="WebVTT cue box">cue box</a> would cover 35% of the video viewport's height.</p>
+ <p>The first cue has its <a lt="WebVTT cue box">cue box</a> positioned at the 10% mark. The
+ "line-left" and "line-right" within the "position" setting indicates which side of the <a
+ lt="WebVTT cue box">cue box</a> the position refers to. Since in this case the text is horizontal,
+ "line-left" refers to the left side of the box, and the cue box is thus positioned between the 10%
+ and the 45% mark of the video viewport's width, probably underneath a speaker on the left of the
+ video image. If the cue was vertical, "line-left" positioning would be from the top of the video
+ viewport's height and the <a lt="WebVTT cue box">cue box</a> would cover 35% of the video
+ viewport's height.</p>
 
  <p>The text within the first cue's cue box is aligned using the "align" cue setting. For
  left-to-right rendered text, "start" alignment is the left of that box, for right-to-left rendered
  text the right of the box. So, independent of the directionality of the text, it will stay
- underneath that speaker. Note that "start" alignment of the cue box is the default for start
- aligned text, so does not need to be specified in "position".</p>
+ underneath that speaker. Note that "center" position alignment of the cue box is the default for
+ start aligned text, in order to avoid having the box move when the base direction of the text
+ changes (from left-to-right to right-to-left or vice versa) as a result of translation.</p>
 
  <p>The second cue has its <a lt="WebVTT cue box">cue box</a> right aligned at the 90% mark of the
- video viewport width ("end" aligned text right aligns the box). The same effect can be achieved
- with "position:55%,start", which explicitly positions the cue box. The third cue has center aligned
- text within the same positioned cue box as the first cue.</p>
+ video viewport width ("right" aligned text right aligns the box). The same effect can be achieved
+ with "position:55%,line-left", which explicitly positions the cue box. The third cue has center
+ aligned text within the same positioned cue box as the first cue.</p>
 
 </div>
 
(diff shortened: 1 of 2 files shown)


zcorpan:
@silviapfeiffer ping

silviapfeiffer:
sorry about the long delay .. looking at it now

index.bs
```diff
@@ -361,13 +361,13 @@ one line except when a line break is definitely necessary.</p>
  <pre>
  WEBVTT
 
- 00:00:00.000 --> 00:00:04.000 position:10%,start align:left size:35%
+ 00:00:00.000 --> 00:00:04.000 position:10%,left align:left size:35%
```
silviapfeiffer: For vertical cues, position refers to a vertical offset. Then, start and end don't refer to left and right, but to top and bottom. I'm reluctant to change these keywords because of that.

zcorpan: I think of it as [line-left](https://drafts.csswg.org/css-writing-modes-3/#line-left), mirroring the behavior of `align:left`. "start" implies different sides for LTR and RTL base direction, which is not the case here.

silviapfeiffer: OK, maybe we should use 'line-left' and 'line-right' then? Those make sense to me.

index.bs
```diff
@@ -375,30 +375,25 @@ one line except when a line break is definitely necessary.</p>
...
+ <p>The first cue has its <a lt="WebVTT cue box">cue box</a> positioned at the 10% mark. The "left"
+ and "right" within the "position" setting indicates which side of the <a lt="WebVTT cue box">cue
+ box</a> the position refers to. Since in this case the text is horizontal, "left" refers to the
  left side of the box, and the cue box is thus positioned between the 10% and the 45% mark of the
  video viewport's width, probably underneath a speaker on the left of the video image. If the cue
- was vertical, "start" positioning would be from the top of the video viewport's height and the <a
+ was vertical, "left" positioning would be from the top of the video viewport's height and the <a
  lt="WebVTT cue box">cue box</a> would cover 35% of the video viewport's height.</p>
 
- <p>The text within the first cue's cue box is aligned using the "align" cue setting. For
```
silviapfeiffer: why did you remove this paragraph?

index.bs
```diff
@@ -728,14 +720,6 @@ consisting of the character U+00C5 LATIN CAPITAL LETTER A WITH RING ABOVE.</p>
   lt="WebVTT cue middle alignment">middle</a>, if there is no <a>WebVTT cue text alignment</a>
   setting for a cue, the <a>WebVTT cue position</a> defaults to 50%.</p>
 
-  <p class="note">Even for <a lt="WebVTT cue horizontal writing direction">horizontal</a> cues with
```
silviapfeiffer: why did you remove this paragraph?

silviapfeiffer:
Feedback from captioners has indicated that for cue boxes that are not explicitly positioned but are smaller than 100% wide, the text alignment needs to determine whether the box is positioned on the left, center or right.

I guess it has to do with the 90% use case: most of the time, left aligned text ends up being on the left edge, right aligned on the right edge, and middle aligned in the center. Only in rare exceptions do you need to place left aligned text at another offset mark.

Therefore, I would recommend to not apply the change to the default positioning that is provided in this patch.

index.bs
zcorpan: OK, sure.

index.bs
zcorpan: I don't remember. I can put it back.

index.bs
zcorpan: I can put it back. Although I disagree with the advice in the last paragraph; see https://www.w3.org/Bugs/Public/show_bug.cgi?id=28257#c12

zcorpan:
Thanks for the review!

> Feedback from captioners has indicated that for cue boxes that are not explicitly positioned but are smaller than 100% wide, the text alignment needs to determine whether the box is positioned on the left, center or right.

This PR still does that for align:left/middle/right.

https://rawgit.com/w3c/webvtt/align-vs-position/index.html#cue-computed-position-alignment

> A WebVTT cue has a computed position alignment whose value is that returned by the following algorithm, which is defined in terms of other aspects of the cue:

>If the WebVTT cue position alignment is not auto, then return the value of the WebVTT cue position alignment and abort these steps.

>If the WebVTT cue text alignment is left, return left and abort these steps.

>If the WebVTT cue text alignment is right, return right and abort these steps.

>Otherwise, return middle.

When you want to align the text to something in the video, you use `align:left` or `align:right` (or `align:middle`), and it does what you want.

What is changed is the position alignment for `align:start` and `align:end`, from line-left and line-right, respectively, to middle. Since `align:start` when it contains RTL text would otherwise end up being positioned on the "wrong" side, and a cue can contain both LTR and RTL text, anything other than middle will be "wrong" for some cases.

zcorpan:
Feedback addressed.

index.bs
```diff
@@ -475,31 +475,33 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+ viewport's height and the <a lt="WebVTT cue box">cue box</a> would cover 35% of the video
+ viewport's height.</p>
 
  <p>The text within the first cue's cue box is aligned using the "align" cue setting. For
  left-to-right rendered text, "start" alignment is the left of that box, for right-to-left rendered
  text the right of the box. So, independent of the directionality of the text, it will stay
- underneath that speaker. Note that "start" alignment of the cue box is the default for start
- aligned text, so does not need to be specified in "position".</p>
+ underneath that speaker. Note that "center" position alignment of the cue box is the default for
+ start aligned text.</p>
```
silviapfeiffer: Maybe add an explicit note that explains why for start/end the "center" position is chosen. Just to make sure we're not confusing people.

index.bs
```diff
@@ -475,31 +475,33 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
 
  <p>The second cue has its <a lt="WebVTT cue box">cue box</a> right aligned at the 90% mark of the
- video viewport width ("end" aligned text right aligns the box). The same effect can be achieved
- with "position:55%,start", which explicitly positions the cue box. The third cue has center aligned
- text within the same positioned cue box as the first cue.</p>
+ video viewport width ("right" aligned text right aligns the box). The same effect can be achieved
```
silviapfeiffer: s/"right" aligned text/"line-right" alignment/

index.bs
```diff
@@ -810,6 +812,11 @@ consisting of the character U+00C5 LATIN CAPITAL LETTER A WITH RING ABOVE.</p>
   automatic position">auto</dfn>, which means the position is to depend on the <a lt="WebVTT cue
   text alignment">text alignment</a> of the cue.</p>
 
+  <p>For <a>WebVTT cues</a> that have a <a lt="WebVTT cue size">size</a> other than 100%, and a <a
+  lt="WebVTT cue text alignment">text alignment</a> of <a lt="WebVTT cue start alignment">start</a>
+  or <a lt="WebVTT cue end alignment">end</a>, authors must not use the default <a lt="WebVTT cue
+  automatic position">auto</a> <a lt="WebVTT cue position">position</a>.</p>
```
silviapfeiffer: Should this be restricted only to cases where "size" is less than 100%? Cause otherwise there is no positioning happening.

Also, would be good to explain why we require explicit positioning - maybe again in a note. Something about the "center" alignment of the text box may confuse captioners that have chosen "start/end" alignment of the text inside the box and feel that the box should be moved to the left or right. However, because of RTL and LTR and mixed directionality text, such automated positioning can lead to unexpected positioning, so therefore we require explicit positioning.

index.bs
```diff
@@ -1792,8 +1792,8 @@ given:</p>
     an optional alignment value consisting of:
     <ol>
      <li>A U+002C COMMA character (,).</li>
-     <li>One of the following strings: "<code>start</code>", "<code>center</code>",
-     "<code>end</code>"</li>
+     <li>One of the following strings: "<code>left</code>", "<code>center</code>",
```
silviapfeiffer: isn't this supposed to be line-left?

index.bs
```diff
@@ -1792,8 +1792,8 @@ given:</p>
...
+     "<code>right</code>"</li>
```
silviapfeiffer: and this should be 'line-right'?

index.bs
```diff
@@ -4361,11 +4361,11 @@ following algorithm.</p>
        direction">horizontal</a></dt>
        <dd>
         <dl class="switch">
-         <dt>If the <a>WebVTT cue line alignment</a> is <a lt="WebVTT cue position center
```
silviapfeiffer: oh good catch

index.bs
```diff
@@ -55,7 +55,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: expose a user interface to the user
             text: text track cue order
         urlPrefix: syntax.html
-            text: syntax-charref
+            text: character references; url: #syntax-charref
```
silviapfeiffer: why mix that into this commit?

silviapfeiffer:
Looks good generally.

index.bs
zcorpan: The "align" setting still uses "left" and "right" keywords.

index.bs
zcorpan: Bikeshed wouldn't build otherwise. It's unrelated so doesn't belong here. I can fix separately and rebase...

index.bs
zcorpan: Added.

index.bs
zcorpan: Yeah, though I'd like to change the internal terms, this is way too error prone. Bikeshed has better ways to do this... :-)

zcorpan:
Thanks; can you also check https://github.com/w3c/webvtt/commit/8b09f98f0598da81ae5974254e16c313c0ba3c15

index.bs
silviapfeiffer: ok, confusion :-)

index.bs
silviapfeiffer: ok, don't worry about it then.

silviapfeiffer:
LGTM

zcorpan:
Thank you!!

silviapfeiffer:
Thanks fort keeping at it - this stuff is non trivial and we're haven't heard much feedback from others.

PR #274, opened 2015-11-30
Use text-wrap:balance

zcorpan:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28432.


silviapfeiffer:
Isn't the paragraph before that also part of the balancing algo?

zcorpan:
No, it's about wrapping. But that should be replaced with overflow-wrap:word-wrap, see

https://www.w3.org/Bugs/Public/show_bug.cgi?id=17218
https://www.w3.org/Bugs/Public/show_bug.cgi?id=17473

silviapfeiffer:
Ok fair enough :-) lgtm

PR #275, opened 2015-11-30
Use overflow-wrap

zcorpan:
Fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=17218.

Also see https://github.com/w3c/webvtt/pull/274 -- if both of these changes are to be merged, the entire list should be removed.


silviapfeiffer:
Wfm

PR #277, opened 2015-12-01
Fix #231: Switch to REGION blocks instead of headers

### LINKED ISSUE #231: Change the syntax of Region headers
See https://www.w3.org/Bugs/Public/show_bug.cgi?id=18657

It is suggested to change the syntax here to use REGION blocks before the cues (consistent with STYLE), instead of headers. I suppose this would also be an opportunity to change from `=` to `:` so that would be consistent with cue settings.

If we use a block it would be possible to allow linebreaks between settings here, if people think that would be nice to be able to do. e.g.

```
WEBVTT

REGION
id:fred
width:40%
lines:3
regionanchor:0%,100%
viewportanchor:10%,90%
scroll:up

REGION
id:bill width:40% lines:3 regionanchor:100%,100% viewportanchor:90%,90% scroll:up

...
```

Apple has shipped support for regions, so need to check with them if this change is OK. @dwsinger ?

zcorpan:
This makes region definitions more consistent with both STYLE
blocks as well as cue settings.

Issue: #231

cc @silviapfeiffer @eric-carlson

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -511,8 +511,22 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
  <pre>
  WEBVTT
- Region: id=fred width=40% lines=3 regionanchor=0%,100% viewportanchor=10%,90% scroll=up
- Region: id=bill width=40% lines=3 regionanchor=100%,100% viewportanchor=90%,90% scroll=up
+
+ REGION
+ id:fred
+ width:40%
+ lines:3
+ regionanchor:0%,100%
+ viewportanchor:10%,90%
+ scroll:up
+
+ REGION
+ id:bill
+ width:40%
+ lines:3
+ regionanchor:100%,100%
+ viewportanchor:90%,90%
+ scroll:up
 
  00:00:00.000 --> 00:00:20.000 region:fred align:left
  &lt;v Fred>Hi, my name is Fred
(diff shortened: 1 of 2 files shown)


dwsinger:
This is not compatible with existing content or implementations, is it?

Do we really need to do it (or do I misunderstand)?

> On Dec 1, 2015, at 3:02 , Simon Pieters <notifications@github.com> wrote:
> 
> This makes region definitions more consistent with both STYLE
> blocks as well as cue settings.
> 
> Issue: #231
> 
> cc @silviapfeiffer @eric-carlson
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/w3c/webvtt/pull/277
> 
> Commit Summary
> 
> 	• Fix #231: Switch to REGION blocks instead of headers
> File Changes
> 
> 	• M index.bs (157)
> 	• M index.html (144)
> Patch Links:
> 
> 	• https://github.com/w3c/webvtt/pull/277.patch
> 	• https://github.com/w3c/webvtt/pull/277.diff
> —
> 

David Singer

silviapfeiffer:
David: at foms I had a chat with Eric about it and he agreed that if we wanted to make it consistent, now would be a good time to do it (Eric should chime in though). It's not a difficult change in the implementations. We're making much more impactful changes elsewhere,

zcorpan:
@dwsinger see https://github.com/w3c/webvtt/issues/231#issuecomment-152547873

zcorpan:
@eric-carlson do you still support this?

eric-carlson:
@zcorpan yes I do. Sorry you had to ping me, I thought I had replied weeks ago.

zcorpan:
Thanks!

PR #279, opened 2015-12-04
Fix #270: Add classes to specify text color

### LINKED ISSUE #270: Add default styling to "well-known" classes
See https://github.com/w3c/webvtt/pull/261#issuecomment-157855519

I think the default rendering should be the same in all user agents, but I don't really mind having a simple default stylesheet for webvtt with some classes like "red".

One problem is if we want to support changing the background color of the background box, we probably have to add a way to specify classes on cues.

This would actually be backwards compatible syntax if we need classes on cues:

```
WEBVTT

00:00:00.000 --> 00:00:10.000.foo.bar align:start
Hello world.
```

### LINKED ISSUE #272: Have tags for specifying color and background color
See https://github.com/w3c/webvtt/pull/261#issuecomment-158187662

Again I suppose this would be needed on the cue level as well as on the element level, if it should be possible to override the default background color.

I don't particularly like adding new elements for color, it seems like it would be better to be able to set colors on any element, like you can set classes on any element.

We could allow color specifiers at the end of start tags, as annotations. For elements that already require annotations, we could make a "#" switch from annotation to color. (It would be possible to write stylesheets that work in legacy clients using `^=` attribute selector, e.g. `v[voice^=Roger]`.)

```
WEBVTT

00:00:00.000 --> 00:00:10.000.foo.bar align:left #FFF #000
<v.foo.bar Roger #FFF #000>Hello
```

The cue itself has classes "foo" and "bar", color "#FFF", background color "#000".
The `v` element has the same classes, annotation "Roger", same colors.

To support non-opaque colors, we can use hex annotation with 4 or 8 characters (this is supported in CSS now, per spec at least). e.g. fully transparent black is #0000 or #00000000. https://drafts.csswg.org/css-color/#hex-notation

The colors specified this way would be overridable with CSS, so it's like HTML `<font color>`, not like `style="..."`.

zcorpan:
Same colors as CEA608, excluding blue and black.

Closes #272.

Issue: #270


silviapfeiffer:
Why go with less colours than cea608?

dwsinger:
> On Dec 4, 2015, at 11:57 , Silvia Pfeiffer <notifications@github.com> wrote:
> 
> Why go with less colours than cea608?

yes, I’d rather match 608 (since 608 colors are referenced by the FCC regulations).

also, double-check, this is using regular styling syntax so CSS-able implementations can use CSS, and hard-wired implentations can hard-wire?

David Singer

silviapfeiffer:
@dwsinger What do you mean by "regular styling syntax"?

If we want to be correct with the mapping of color names of CEA608 to WebVTT, we have to use "green" and map it to "lime" (as Simon has done here and as the mapping spec does, too https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html).

If instead you want to use the CSS named colors, we will need to explain to use "lime" for the CEA608 "green". I wouldn't want to do that, because this isn't CSS.

dwsinger:
@silvia i mean, marked up in a way that CSS, if available, can be used to define the colors. don't invent new markup which can't be supported by CSS. No, I don't mind the divergence from CSS named colors. I just don't want implementations that have CSS, to have to support colors separately/explicitly as well as supporting CSS.

zcorpan:
A CSS UA would totally use CSS to implement this; it's defined in terms of CSS.

dwsinger:
> On Dec 7, 2015, at 12:21 , Simon Pieters <notifications@github.com> wrote:
> 
> A CSS UA would totally use CSS to implement this; it's defined in terms of CSS.

It was probably blindingly obvious that this had to be true, but I wanted to state it for the record.


David Singer

tairt:
I will try to get more feedback for this merge proposal.

tairt:
I reached out to some partners in the HBB4ALL project and so far I did not get any objection. For me the commit seems to me like a reasonable basic solution that covers the minimal requirements regarding the use of text color.

I got the following feedback from two Spanish partners. One of them analysed the the [AENOR Norm UNE 153010:2012 Subtitulado para personas sordas y personas con discapacidad auditiva](http://www.aenor.es/aenor/normas/normas/fichanorma.asp?tipo=N&codigo=N0049426&PDF=Si#.Vmrher9i-6F) in respect to the proposal: 


> In principle, the solution you propose fulfils the requirements of the Spanish code of good practice. The colour is used to distinguish the characters in Spain, as said. This code does not contain a list of possible colours, but a rule to ensure two colours are different enough. The rule is the "colour difference", evaluated as the sum of the absolute value of the differences in each RGB value. If you consider two colours, identified by RGB values (8 bits) (R1G1B1 and R2G2B2), you calculate the "colour difference" as |R1-R2| + |G1-G2| + |B1-B2|. According to the code, this result must be at least 255 to ensure the colours are different enough. The colours you propose fulfil this requirement, although I am not sure in the case of the red.

> Regarding to the background, a box is not mandatory. The only requirement is to keep a contrast ratio of at least 4,5 between text and box/contour. And this "contrast ratio" is calculated according to WCAG 2.0 (http://www.w3.org/TR/WCAG20/#contrast-ratiodef). I am not sure, but an opacity of 0.8 may ensure a good contrast whatever the colour of the video at that moment behind the box.

> With regards to blue, it was used for the lyrics of a song (on yellow background in the case of teletext) but this just applies to "legacy subtitles".

> So, I think your basic solution is globally valid according to this code.

The other spanish partner mentioned:

> I would try to mention that red should be avoided.

From a German broadcaster I got the following feedback:

> The proposal looks fine to me. It make sense to me to disallow blue when the background color is black. If blue would be allowed there should also be the possibility to set the background color.

dwsinger:
> On Dec 11, 2015, at 7:03 , Andreas Tai <notifications@github.com> wrote:
> 
> I reached out to some partners in the HBB4ALL project and so far I did not get any objection. For me the commit seems to me like a reasonable basic solution that covers the minimal requirements regarding the use of text color.
> 
> I got the following feedback from two Spanish partners. One of them analysed the the AENOR Norm UNE 153010:2012 Subtitulado para personas sordas y personas con discapacidad auditiva in respect to the proposal:
> 
> In principle, the solution you propose fulfils the requirements of the Spanish code of good practice. The colour is used to distinguish the characters in Spain, as said. This code does not contain a list of possible colours, but a rule to ensure two colours are different enough. The rule is the "colour difference", evaluated as the sum of the absolute value of the differences in each RGB value. If you consider two colours, identified by RGB values (8 bits) (R1G1B1 and R2G2B2), you calculate the "colour difference" as |R1-R2| + |G1-G2| + |B1-B2|. According to the code, this result must be at least 255 to ensure the colours are different enough. The colours you propose fulfil this requirement, although I am not sure in the case of the red.
> 
> Regarding to the background, a box is not mandatory. The only requirement is to keep a contrast ratio of at least 4,5 between text and box/contour. And this "contrast ratio" is calculated according to WCAG 2.0 (http://www.w3.org/TR/WCAG20/#contrast-ratiodef). I am not sure, but an opacity of 0.8 may ensure a good contrast whatever the colour of the video at that moment behind the box.
> 
> With regards to blue, it was used for the lyrics of a song (on yellow background in the case of teletext) but this just applies to "legacy subtitles".
> 
> So, I think your basic solution is globally valid according to this code.
> 
> The other spanish partner mentioned:
> 
> I would try to mention that red should be avoided.
> 
> From a German broadcaster I got the following feedback:
> 
> The proposal looks fine to me. It make sense to me to disallow blue when the background color is black. If blue would be allowed there should also be the possibility to set the background color.

Great research, thanks!

so, the idea is to say that a user-agent not supporting CSS must nonetheless support coloring via a set of named classes which are (all of | a subset of) the color classes from the x08 styles, and should also allow the user to over-ride what color is shown for a given class. 

oh, and VTT files using those classes must, of course, explicitly include a style-sheet defining them, for the CSS-aware UAs.

??

David Singer

tairt:
@dwsinger Yes, that is my understanding.

zcorpan:
> oh, and VTT files using those classes must, of course, explicitly include a style-sheet defining them, for the CSS-aware UAs.

No, that would not be necessary. CSS-aware UAs would also have these as default styles, like you don't need to explicitly specify that `<i>` should be italic.

zcorpan:
Thank you @tairt!

OK so it is good to exclude blue, so long as we don't support changing the background.

What about red?

> The other spanish partner mentioned:
>
>> I would try to mention that red should be avoided.

In https://github.com/w3c/webvtt/pull/279#issuecomment-163958038 I noted that red on \#333333 is only 3.16:1 in contrast, although red on \#000000 is 5.25:1.

I suppose red could be used if it's used together with bold, where the WCAG contrast requirement for "large text" could apply (where 3:1 OK).

foolip:
Has any browser implementor expressed interest in implementing something like this? I struggle to see a reason to implement this in Chromium, other than interoperability if everyone else implements it.

zcorpan:
> The rule is the "colour difference", evaluated as the sum of the absolute value of the differences in each RGB value. If you consider two colours, identified by RGB values (8 bits) (R1G1B1 and R2G2B2), you calculate the "colour difference" as |R1-R2| + |G1-G2| + |B1-B2|. According to the code, this result must be at least 255 to ensure the colours are different enough. The colours you propose fulfil this requirement, although I am not sure in the case of the red.

They all do, including red. Although I think it is bogus to calculate color difference like that. See https://en.wikipedia.org/wiki/HSL_and_HSV#Motivation -- take for example \#ffff00 and \#bbbb77, where the sRGB difference is 255, but the contrast ratio is only 1.86:1.

In http://webaim.org/blog/wcag-2-0-and-link-colors/ I find that WCAG 2.0 requires a 3:1 contrast between normal text and link text. It seems like the same should apply here.

zcorpan:
@foolip no. It is indeed pretty unnecessary if CSS is supported, but it appears that not all WebVTT implementations have CSS and some color support is still a requirement, and I think it would be good to have the same rendering in CSS-aware as CSS-unaware UAs (when there's no author-level CSS).

dwsinger:
Sent from my iPad

> On Dec 14, 2015, at 5:51 AM, Simon Pieters <notifications@github.com> wrote:
> 
> oh, and VTT files using those classes must, of course, explicitly include a style-sheet defining them, for the CSS-aware UAs.
> 
> No, that would not be necessary. CSS-aware UAs would also have these as default styles, like you don't need to explicitly specify that <i> should be italic.
> 


Ok, you want that true for the colors also? So there are some default styles, colors etc. built in?

dwsinger:
> On Dec 14, 2015, at 6:27 , Simon Pieters <notifications@github.com> wrote:
> 
> Thank you @tairt!
> 
> OK so it is good to exclude blue, so long as we don't support changing the background.
> 
> What about red?
> 
> The other spanish partner mentioned:
> 
> I would try to mention that red should be avoided.
> 
> In #279 (comment) I noted that red on #333333 is only 3.16:1 in contrast, although red on #000000 is 5.25:1.
> 
> I suppose red could be used if it's used together with bold, where the WCAG contrast requirement for "large text" could apply (where 3:1 OK).

I really think we should pick an existing defined set of colors, and not try to micro-manage it smaller. The x08 set seems best.


David Singer

zcorpan:
> Ok, you want that true for the colors also?

Yes

zcorpan:
> I really think we should pick an existing defined set of colors, and not try to micro-manage it smaller.

Why? Nobody wants blue or black, as far as I can tell. I think it makes no sense to have a language feature that makes text impossible to read.

https://zcorpan.github.io/live-webvtt-viewer/#style=::cue+{+color:blue+}

If we add a way to change or turn off the background, I'm OK with adding blue and black, but not now.

dwsinger:
> On Dec 14, 2015, at 11:05 , Simon Pieters <notifications@github.com> wrote:
> 
> I really think we should pick an existing defined set of colors, and not try to micro-manage it smaller.
> 
> Why?

a) we don’t need a long discussion over the benefits of various colours
b) we can claim to support the colors and the standards mentioned in the regulations, with a straight face

I should check the regulations.
 
> Nobody wants blue or black, as far as I can tell. I think it makes no sense to have a language feature that makes text impossible to read.
> 
> https://zcorpan.github.io/live-webvtt-viewer/#style=::cue+{+color:blue+}
> 
> If we add a way to change or turn off the background, I'm OK with adding blue and black, but not now.

Sure, some of these colors make no sense unless the background is controlled or known (black makes perfect sense if the video is know to be predominantly very light).

David Singer

dwsinger:
(6) Caption background color and opacity. All apparatus shall implement captioning such that the caption background may be displayed in the 64 colors defined in CEA-708 and such that users are provided with the ability to override the authored color for the caption background and select from a palette of at least 8 colors including: white, black, red, green, blue, yellow, magenta, and cyan. All apparatus shall implement captioning such that users are provided with the ability to vary the opacity of the caption background and select between opaque, semi-transparent, and transparent background opacities.
…
(8) Caption window color. All apparatus shall implement captioning such that the caption window color may be displayed in the 64 colors defined in CEA-708 and such that users are provided with the ability to override the authored color for the caption window and select from a palette of at least 8 colors including: white, black, red, green, blue, yellow, magenta, and cyan. All apparatus shall implement captioning such that users are provided with the ability to vary the opacity of the caption window and select between opaque, semi-transparent, and transparent background opacities.


David Singer

zcorpan:
No, webvtt cues have a background of rgba(0,0,0,0.8), i.e. black with 20% opacity. It is always dark. This PR doesn't provide any way to turn it off.

zcorpan:
My goal was to address the feedback collected by @tairt, not to match regulations. In what you quote alone, this PR is not enough even if blue and black are added. And there are lots of other things in there, if we are to support everything we are reinventing CSS, and then it makes more sense to require support for CSS instead imo.

silviapfeiffer:
@zcorpan the quoted regulations require 8 background colours and 8 foreground colours and 3 opacity states. That's hardly reinventing CSS. The regulations incorporate experience just as much as an experience report from users, so I wouldn't dismiss them out of hand.

zcorpan:
> the quoted regulations require 8 background colours and 8 foreground colours and 3 opacity states.

It requires 64 colors for apparatus. But even so, this PR doesn't support changing the background, so my statement stands.

> That's hardly reinventing CSS.

I mean if we are to add everything the regulations require, not jut colors. I thought there was more to it?

tairt:
I agree with the position of @zcorpan. If we do not add more features as requirements this pull request is a resonable solution.

The regulation copied in by @dwsinger makes a customization features mandatory. If this is goes into WebVTT this of course changes the discussion.

Regarding CEA608 and 708 regulation in North America: I do not say this should be ignored but I think it only one region should be the reference for a standard that should be applied globally. The comments I posted also reflects feedback of five other different countries in Europe and I think it is good to capture as many requirements from daily operations as possible. 

As @zcorpan states it makes no sense to allow black on black or blue on black. This is indeed a real problem and you will see on some broadcast services that teletext subs with teletext colors get converted to DVB Bitmap subtitles where the background is the same as now the default WebVTT background (black with 0.8 opacity). Unfortunately blue is not excluded and these subtitles are now unreadable (fortunately the color combination blue on white is often only used for naming the author).

Although I am fine with this proposal the following could be a way out: a) accept this pull request as a starting point b) adding a new issue to allow background colors c) add more text colors when a pull request adds classes for background colors.

tairt:
Regarding the question about the color red for captioning I got more feedback. Personally I would not object in allowing red but there could be some more information.

Some views:

_Spain_
"Red is a problematic colour for almost 10% of men who are colour blind.
Also for this reason green and blue are not used, but cyan."

_Switzerland_
"Red has not been used for subtitling for already quite some time. Especially the red of teletext is very dark and it is very hard to read against a black background."

_UK_
"See http://jfly.iam.u-tokyo.ac.jp/color/index.html
It’s difficult to find good colours, but certainly monochromatic colours (red, green and blue) are less readable to the colour blind population than the ‘redundant coded’ colours (yellow, cyan, magenta)."

dwsinger:
> On Dec 15, 2015, at 09:35, Andreas Tai <notifications@github.com> wrote:
> 
> @zcorpan <https://github.com/zcorpan> Regarding the question about the color red for captioning I got more feedback. Personally I would not object in allowing red but there could be some more informal information.
> 
> Some views:
> 
> Spain
> "Red is a problematic colour for almost 10% of men who are colour blind.
> Also for this reason green and blue are not used, but cyan."
> 
> Switzerland
> "Red has not been used for subtitling for already quite some time. Especially the red of teletext is very dark and it is very hard to read against a black background."
> 
> UK
> "See http://jfly.iam.u-tokyo.ac.jp/color/index.html <http://jfly.iam.u-tokyo.ac.jp/color/index.html>
> It’s difficult to find good colours, but certainly monochromatic colours (red, green and blue) are less readable to the colour blind population than the ‘redundant coded’ colours (yellow, cyan, magenta).”
> 
Us mentioning a color doesn’t mean that authors have to use it. There are all sorts of color perception problems — red/green is common but by no mean the only one (actually, it’s deficiencies or color shifts in up to 3 sensors, but…)

silviapfeiffer:
> Although I am fine with this proposal the following could be a way out: a) accept this pull request as a starting point b) adding a new issue to allow background colors c) add more text colors when a pull request adds classes for background colors.

That works for me.

zcorpan:
@rgov is there interest in implementing this for Safari?

zcorpan:
@alastor0325 is there interest in implementing this for Gecko?

zcorpan:
It appears this has no interest to implement in browser vendors. Closing.

dwsinger:
This bug had a conclusion; I did not notice the questions going to people I don't know. I still think we should add the fixed x08 colors, as apparently agreed.

zcorpan:
@dwsinger will it be implemented in Safari?

dwsinger:
> On Jan 19, 2017, at 15:43 , Simon Pieters <notifications@github.com> wrote:
> 
> @dwsinger will it be implemented in Safari?

I can’t promise the future, but I will check on the present…

> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

David Singer

silviapfeiffer:
Note that this bug was not actually targeted at browser vendors, but at WebVTT implementations that don't support CSS. e.g. what would vlc do to get colors? For browser, we can always say that if you need these colors, include the style sheet.

zcorpan:
That would set a precedent in diverging behavior between browsers and non-browser implementations. I would like WebVTT implementations to be interoperable.

dwsinger:
I completely agree that they should be interoperable. The suggestion was to define these as classes that the VTT client would have pre-defined either (a) as a style-sheet pre-loaded into the UA, if CSS is supported, or (b) 'as if' that style-sheet were loaded, i.e. hard-wired, but under the same syntax. I.e. we'd write in the spec. "The following styles must be defined as part of VTT client initialization, either by loading this style sheet, or by 'hard wiring' in the UA."  (Roughly, word-smithing needed).

dwsinger:
I think (I don't know) that if they had been specified at the time that our non-browser implementation was written, we would have implemented. So, it's something I can't rule out in an update. I still feel we could say that implementations must behave "as if the following CSS stylesheet were preloaded" (which browser implementations can easily achieve by, well, pre-loading).

silviapfeiffer:
I still stand behind https://github.com/w3c/webvtt/pull/279#issuecomment-165274626

dwsinger:
sure, let's find a consensus that all UAs (browser-based, and non-browser) can support; mine was only one solution. I certainly agree that we don't want a situation where the two classes of UA behave differently (as I obviously implied by suggesting the as-if-CSS-loaded approach, but there are others).

silviapfeiffer:
What I meant was that I'd like colors to go it, whether we start with a smaller set or the full, I am not so fussed about. :-)

dwsinger:
Agree with Silvia, and I am not fussed about how we achieve CSS and non-CSS equivalence, either

silviapfeiffer:
actually going to reject this one because we're dealing with it in #395

Issue #310, opened 2016-07-01
Parser drops identifier on the floor


zcorpan:
@alastor0325 noticed that the spec's parser doesn't set the cue identifier to anything but the empty string. Oops. This regressed in https://github.com/w3c/webvtt/pull/246

Fix is to set the cue identifier to *buffer* on cue creation in the parser.

alastor0325:
Before to set the cue identifier to `buffer` on cue creation in the parser,  whether there is another mistake that can't handle ` -->` correctly for **cue identifier**?

Could you help me to verify the following steps? 

Take [this file](http://w3c-test.org/webvtt/webvtt-file-format-parsing/webvtt-file-parsing/support/arrows.vtt) as example, 

In step 13-1 in **WebVTT parser algorithm**, we would go to read the the line ` -->`. 
> Collect a WebVTT block, and let block be the returned value.

Since `line` contains "-->", then go to the step 11-4 in **collect a WebVTT block**
> If line contains the three-character substring "-->" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN), then run these substeps:

In step 11-4-1-4, run **Collect WebVTT cue timings and settings**, but it would fails because the line doesn't have any timestamp.
> Collect WebVTT cue timings and settings from line using regions for cue. If that fails, let cue be null. Otherwise, let buffer be the empty string.

After above steps, we parse identifier fail, so we would get next line to `line`. Since the `line` still contains  "-->", go to the step 11-4 in **collect a WebVTT block** again.

However, `seen arrow` has already be set to true in previous steps, so we won't go to step 11-4-1 
> line count is 2 and seen arrow is false

We would go here, and break out the loop. At that time, the `cue` is null and the `position` is `previous postion`, that means it points back to the cue identifier place of the file. (string ` -->`)
> Otherwise, let position be previous position and break out of loop.

Therefore, go back to  step 13-1 in **WebVTT parser algorithm**. Since we don't get the valid value for `block`, we go to next loop and the result would still be the same. (doesn't get the cue)

zcorpan:
I think your analysis is correct except for this:

> At that time, the cue is null and the position is previous postion, that means it points back to the cue identifier place of the file. (string `-->`)

In particular consider steps 11-3 and 11-4-1-2 of **collect a WebVTT block** in the first iteration. *previous position* will point to the start of the line `00:00:00.000 --> 00:00:01.000`.

PR #312, opened 2016-07-06
Modify the parsing rule to set the cue's identifier.

alastor0325:
According to #310, the parser doesn't set the cue identifier to anything but the empty string. 
The fix is to set the cue identifier to buffer on cue creation.

@zcorpan, could you help me review this patch?
Thanks!

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2194,7 +2194,7 @@ header</i> set, the user agent must run the following steps:</p>
 
         <ol>
 
-         <li><p>Let |cue|'s <a>text track cue identifier</a> be the empty string.</p></li>
+         <li><p>Let |cue|'s <a>text track cue identifier</a> be |buffer|.</p></li>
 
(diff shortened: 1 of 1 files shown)


zcorpan:
Change looks good, but you should edit `index.bs` instead.

You also should do something to make the **ipr** checker happy. cc @plehegar

alastor0325:
Hi, @zcorpan,
Could you help me review the patch again?
In addition, the ipr checker said that I'm not in the right group, which group I should join in?
Thanks!

zcorpan:
Looks good!

https://www.w3.org/AudioVideo/TT/#join

alastor0325:
hi, @zcorpan,
sorry for this stupid question,
I have already joined that group, and what's the next steps to make my commit to pass the checker?
Thanks!

zcorpan:
Great! Try force-pushing an amended commit to rerun the checker.

alastor0325:
Hi, @zcorpan,
Do I still need to do any modification? or we can merge this commit?
Thanks :)

PR #321, opened 2016-10-15
Separated interpretation of cue settings into their correct sections.

### LINKED ISSUE #315: CR bug: influence of writing direction on line, position and size
https://www.w3.org/Bugs/Public/show_bug.cgi?id=28071

We think we should replace this normative text by a note that says "Note that writing direction affects the interpretation of line, position, and size" and then these two paragraphs should be placed correctly to define position, line and size each, as they *use* writing direction to give meaning to a value, i.e. they are not about writing direction per se.

silviapfeiffer:
This fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=28071.
    Also fixes #315 .

Note that I ran format.py on the source file and it found some
    issues with missing end tags, which I fixed. It also rearranged some text.


silviapfeiffer:
Marked as non-substantive for IPR from ash-nazg.

.htaccess
```diff
@@ -0,0 +1 @@
+Redirect 301 / https://w3c.github.io/webvtt/
```
zcorpan: Was the `.htaccess` intentional?

zcorpan:
Oh maybe you branched off `master` instead of `gh-pages`?

zcorpan:
I cherry-picked https://github.com/w3c/webvtt/pull/321/commits/64a0a79a49c5851b70690f697975bc5606fe36f6 and landed as https://github.com/w3c/webvtt/commit/039455f57177275b72baada84c1faab98140e1f2

Thank you!

silviapfeiffer:
I'm not sure what happened - I did branch off gh-pages, but then this appeared. Thanks for the cherry pick!

Issue #325, opened 2017-02-06
Wrong name in 'collect WebVTT region settings'


BenjaminSchaaf:
Under [5.2.](https://w3c.github.io/webvtt/#region-settings-parsing) step 2.4. match for `viewportanchor`:

> 5. Let region’s `WebVTT region anchor point` be the tuple of the percentage values calculated from *viewportanchorX* and *viewportanchorY*.

`WebVTT region anchor point` should read `WebVTT region viewport anchor point`.

It has the correct link, but the text is wrong.

PR #326, opened 2017-02-06
Fix for #325

BenjaminSchaaf:
Under 5.2. Step 2.4. match for viewportanchor, Step 5.: Update `WebVTT region anchor point` text to match link.


zcorpan:
Marked as non-substantive for IPR from ash-nazg.

zcorpan:
Thank you!

PR #328, opened 2017-02-09
Add an EOF check before header

### LINKED ISSUE #327: Missing EOF check in 'WebVTT parser algorithm'
In [5.1.](https://w3c.github.io/webvtt/#file-parsing) 'WebVTT parser algorithm', Step 9 may advance the `position` possibly leading to an EOF while Step 10 assumes no EOF has been reached.

If at Step 9 the character indicated by `position` is a LF followed by the EOF a direct translation of the algorithm may crash.

zcorpan:
Fixes #327.

---

Thanks @BenjaminSchaaf; please take a look.


BenjaminSchaaf:
Looks good

PR #332, opened 2017-02-21
Set |seen cue| to true when seeing a cue

### LINKED ISSUE #324: Unused Variable 'seen cue'
Under 'WebVTT parser algorithm' (3.) `seen cue` is set to false, and is then only checked.

Should this variable be removed? (can make a pr if so)

zcorpan:
This variable was intended to ignore STYLE and REGION blocks after
having seen a cue, but it was never set to true.

Fixes #324.


silviapfeiffer:
LGTM

PR #334, opened 2017-02-22
Throw IndexSizeError if VTTRegion lines is set to -1

### LINKED ISSUE #330: VTTRegion API Overview regionAnchorY not defined
In the overview of the [VTTRegion interface](https://w3c.github.io/webvtt/#the-vttregion-interface) (underneath in the interface definition), the second `regionAnchorX` should be named `regionAnchorY`.

### LINKED ISSUE #333: Setting VTTRegion.lines to a negative value should throw
http://w3c.github.io/webvtt/#the-vttregion-interface

> The lines attribute, on getting, must return the WebVTT region lines of the WebVTT region that the VTTRegion object represents\. On setting, the WebVTT region lines must be set to the new value\.

Test: http://w3c-test.org/webvtt/api/VTTRegion/lines.html

In Chrome 58.0.3020.0 canary (which has this fix https://bugs.chromium.org/p/chromium/issues/detail?id=690014#c3 ) this throws IndexSizeError if value is < 0.

https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/html/track/vtt/VTTRegion.cpp?l=111&rcl=861e1e50380132d7028812bba3f3d4f073fa4c26

The syntax does not support negative values, so it would be sensible to throw for negative values. I suggest we change the spec and test to match Chromium implementation.

cc @fsoder @BenjaminSchaaf

zcorpan:
Fixes #333. Also drive-by fix #330.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5533,7 +5533,8 @@ interface VTTRegion {
 
  <dt>|region| . {{VTTRegion/lines}}</dt>
  <dd>
-  <p>Returns the text track region height as a number of lines. Can be set.</p>
+  <p>Returns the text track region height as a number of lines. Can be set. Throws an
+  {{IndexSizeError}} if the new value is negative.</p>
  </dd>
 
  <dt>|region| . {{VTTRegion/regionAnchorX}}</dt>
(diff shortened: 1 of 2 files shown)


BenjaminSchaaf:
LGTM

PR #337, opened 2017-03-03
Drop line: and position: settings with invalid linealign/colalign

### LINKED ISSUE #336: Parsing: should drop line:100,invalid setting
http://w3c.github.io/webvtt/#cue-timings-and-settings-parsing

> Let cue’s WebVTT cue line be number\.
> 
> If the last character in linepos is a U\+0025 PERCENT SIGN character \(%\), then let cue’s WebVTT cue snap\-to\-lines flag be false\. Otherwise, let it be true\.

I think we should validate `linealign` before setting the WebVTT cue line. Equivalent for position setting. It is bad for future extensibility to use the line/position with unrecognized alignment.

(Noticed this when looking at @BenjaminSchaaf's revamp of parsing tests.)

zcorpan:
Fixes #336.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2639,8 +2639,7 @@ run the following steps:</p>
        and let |linealign| be the trailing substring of |value| starting from the character
        immediately after the first U+002C COMMA character (,) in that string.</p></li>
 
-       <li><p>Otherwise let |linepos| be the full |value| string and |linealign| be the empty
-       string.</p></li>
+       <li><p>Otherwise let |linepos| be the full |value| string and |linealign| be null.</p></li>
 
        <li><p>If |linepos| does not contain at least one <a lt="ASCII digits">ASCII digit</a>, then
        jump to the step labeled <i>next setting</i>.</p></li>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -2632,10 +2632,11 @@ run the following steps:</p>
 
       <ol>
 
-       <li><p>If |value| contains a U+002C COMMA character (,), then let |linepos| be the leading
-       substring of |value| up to and excluding the first U+002C COMMA character (,) in that string
-       and let |linealign| be the trailing substring of |value| starting from the character
-       immediately after the first U+002C COMMA character (,) in that string.</p></li>
+       <li><p>If |value| contains a U+002C COMMA character (,) that is not the last character in
```
BenjaminSchaaf: What's the reason behind making this more complicated? The substring after the `,` when its at the end should be empty anyway, right?

BenjaminSchaaf:
LGTM otherwise.

Other than the step 1 changes, I've already changed the parser in wpt to match, pending changes to other tests.

index.bs
zcorpan: I wanted `line:50,` to be dropped. But maybe can set linealign to null instead of empty string, and check for that, instead.

silviapfeiffer: If we want the comma version to take priority over the non comma version, we should allow this as one where the parameters take the default value, I.e. linealign=start, seeing as if we make this extensible with further parameters, we will not always expect all the parameters to be present.
E.g. line:50,XYZ=blah should still take precedent over line:50 and result in linealign getting the default value.

zcorpan: No. ^_^ This is ugly as it is, let's not consider making it worse. But even so, XYZ is invalid now and should still cause the whole thing to be dropped.

zcorpan:
Fixed step 1, PTAL

silviapfeiffer:
WFM

zcorpan:
@BenjaminSchaaf 

> Other than the step 1 changes, I've already changed the parser in wpt to match, pending changes to other tests.

Could you upload this as a PR to web-platform-tests, even if incomplete?

BenjaminSchaaf:
Created here: https://github.com/w3c/web-platform-tests/pull/5113

zcorpan:
Bug for Gecko:
https://bugzilla.mozilla.org/show_bug.cgi?id=1346185

Issue #307, opened 2016-06-23
Linepos can be the floating value


alastor0325:
From [comment](https://github.com/w3c/web-platform-tests/issues/3222#issuecomment-227711766) in web-platform-tests.

> If linepos contains any characters other than U+002D HYPHEN-MINUS characters (-) and ASCII digits, then jump to the step labeled next setting.

We should make the parser rule and DOM API consistent, to accept the floating value when parsing the **linepos** value.

zcorpan:
I'm OK with this. But I think we should keep the *conforming* syntax as is -- it is an authoring mistake to not use an integer here.

alastor0325:
Sorry, I don't understand your comment very well :(

> I think we should keep the conforming syntax as is -- it is an authoring mistake to not use an integer here.

Does it mean we should keep the original description or ....? 

Thanks!

zcorpan:
I mean I'm OK with changing the parser rules to accept `line:1.5`, but we should keep the current syntax rules so that a webvtt validator will complain about it.

alastor0325:
So... you means we shouldn't modify the spec now, but change the test of the web-platform-test to allow `line:1.5`.
Do I understand correctly?
Thanks!

zcorpan:
No. 😊

The spec has two sections:
https://w3c.github.io/webvtt/#syntax - this defines how to *write* a *conforming* WebVTT file.
https://w3c.github.io/webvtt/#parsing - this defines how to *parse* **anything** as WebVTT (whether it is conforming or not).

There is no implied relationship between what is conforming syntax and how you are required to parse it. (c.f. lack of whitespace between cues.)

I am OK with changing the parsing rules (and the tests), but not the syntax section.

I don't know how to make this clearer. Feel free to jump in on [IRC](https://wiki.whatwg.org/wiki/IRC) if you want to chat.

alastor0325:
Thanks for explanation!!!

I were confused because the syntax part about `line` doesn't need to be changed since it has already allowed any floating value, is that right?

> Or to represent a line number  
1.Optionally a U+002D HYPHEN-MINUS character (-).
2.One or more ASCII digits. 

Therefore, if we could change the parsing rule, could you help me to modify it?
And I would send PR after I modified the test.

Thanks again :)

zcorpan:
ASCII digits is 0-9 but does not include "."

alastor0325:
Ok, Thanks!
Close this bug because it would be fixed in #308.

[...]

PR #339, opened 2017-03-03
Use HTML's rules for parsing floating-point number values

### LINKED ISSUE #338: "Interpret linepos as a (potentially signed) rea..."
http://w3c.github.io/webvtt/#cue-timings-and-settings-parsing

> Interpret linepos as a \(potentially signed\) real number, and let number be that number\.

Unclear what this means for `-0` or if the output can be Infinity/-Infinity.

I think we should use https://html.spec.whatwg.org/multipage/infrastructure.html#rules-for-parsing-floating-point-number-values which parses `-0` into +0 and avoids infinities.

cc @fsoder

zcorpan:
This makes line:-0 parse into +0 and ignores the setting if is
too big to represent as a finite number in IEEE 754.

Fixes #338.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2511,10 +2511,12 @@ means that it is aborted at that point and returns nothing.</p>
 
  <li><p>If |input| does not match the syntax for a <a>WebVTT percentage</a>, then fail.</p></li>
 
- <li><p>Ignoring the trailing percent sign, interpret |input| as a real number. Let that number be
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
Hmm, I think we want integers and not floats?

zcorpan:
Nope, see https://github.com/w3c/webvtt/issues/307

Issue #264, opened 2015-11-16
incomplete ruby implementation [I18N-ISSUE-431]


r12a:
[moving to github from https://www.w3.org/Bugs/Public/show_bug.cgi?id=28265]

i was just reviewing this with a mind to close our i18n tracker issue for now (and reopen as support for the other aspects of the html5 markup model spreads), when it occurred to me that webvtt doesn't support the `rb` element.  

This is widely supported in browsers (see the test results at http://www.w3.org/International/tests/repo/results/ruby-html#position), and i think it should therefore be supported by webvtt.  Adding it allows for additional styling options, as well as reducing the potential for confusion in authors, who are used to using it in HTML.

Note that HTML5 supports all of the following:

```
<ruby><rb>...</rb><rt>...</rt>...
<ruby><rb>...<rt>...
<ruby>...<rt>...</rt>...
```

zcorpan:
In WHATWG HTML, `ruby`, `rt` and `rp` are conforming, the others are not.  `rb` doesn't have any special styling by default in WebKit and Blink. In particular see http://software.hixie.ch/utilities/js/live-dom-viewer/saved/3749

To style the ruby base in VTT, you can use the `<c>` element (or I suppose in most cases you can just style the `<ruby>`). Similarly in HTML, you can use a `<span>`.

I think adding `rb` to VTT should wait until all browsers have adopted the necessary ruby styling for HTML.

I think we could add `rp` now though. Since it appears that ruby is not supported in some VTT implementations, it might be desirable to have fallback parens there. The only issue would be legacy ruby-supporting clients that don't support `rp` in VTT, would render the parens also.

zcorpan:
> The only issue would be legacy ruby-supporting clients that don't support `rp` in VTT, would render the parens also.

We *could* use markup like

    <ruby>base<(><rt>text<)></ruby>

but it would be different from HTML and it wouldn't render the parens in existing ruby-unsupporting VTT implementations. It would also only allow simple parens.

silviapfeiffer:
I wouldn't complicate things with special parens tags - I'd expect browsers will implement uniform ruby support instead. Also, I agree with waiting with `<rb>` - it seems it's only used for styling and there's no support in most browsers for that yet.

dwsinger:
We should re-assess the status of Ruby implementations in all browsers and consider whether we need to adapt VTT in the light of current practice.  Re-opening.

silviapfeiffer:
[...]

PR #347, opened 2017-05-06
Ruby issue264

silviapfeiffer:
Include rb as a ruby tag in WebVTT.

    * added rb tag in syntax, parsing and rendering
    * also ran format.py, which re-wrapped a few other paragraphs

Addresses issue #264


index.bs
```diff
@@ -1491,13 +1495,25 @@ underlined text, and a <a>WebVTT cue span end tag</a> "<code>u</code>".</p>
...
+   </ol>
+   <li>a ruby base text component:</li>
+   <ol>
+    <li>A <a>WebVTT cue span start tag</a> "<code>rb</code>" that disallows an annotation.</li>
+    <li>A <dfn>WebVTT cue ruby base text span</dfn>: <a>WebVTT cue internal text</a>, representing
+    the ruby base text component of the ruby annotation.</li>
+    <li>A <a>WebVTT cue span end tag</a> "<code>rb</code>". If this is the last occurrence of this
+    group of components in the <a>WebVTT cue ruby span</a> or a <a>WebVTT cue span start tag</a>
+    "<code>rt</code>" is next, then this last end tag string may be omitted.</li>
+   </ol>
```
zcorpan: This appears to be wrong... It allows `<ruby>foo<rt>bar<rb>baz</ruby>`, but not `<ruby><rb>foo<rt>bar</ruby>`

index.bs
```diff
@@ -3092,6 +3118,12 @@ string |input| supposedly containing <a>WebVTT cue text</a>, and optionally a fa
       Object">attach</a> a <a>WebVTT Ruby Text Object</a>.</p>
      </dd>
 
+     <dt>If the tag name is "<code>rb</code>"</dt>
+     <dd>
+      <p>If |current| is a <a>WebVTT Ruby Object</a>, then <a lt="attach a WebVTT Internal Node
```
zcorpan: Should this pop an `rt`? (And vice versa?)

index.bs
```diff
@@ -4546,17 +4587,21 @@ corresponding <a lt="text track cue">cue</a>'s <a>WebVTT cue text alignment</a>:
 ''rgba(255,255,255,1)''. [[!CSS3-COLOR]]</p>
 
 <p>The 'background' shorthand property on the <a>WebVTT cue background box</a> and on <a>WebVTT Ruby
-Text Objects</a> must be set to ''rgba(0,0,0,0.8)''. [[!CSS3-COLOR]]</p>
+Text Objects</a> and <a>WebVTT Ruby Base Text Objects</a> must be set to ''rgba(0,0,0,0.8)''.
```
zcorpan: No, not `rb` objects here.

index.bs
```diff
@@ -4546,17 +4587,21 @@ corresponding <a lt="text track cue">cue</a>'s <a>WebVTT cue text alignment</a>:
...
+[[!CSS3-COLOR]]</p>
 
 <p>The 'white-space' property on the (root) <a>list of WebVTT Node Objects</a> must be set to
-''white-space/pre-line''. [[!CSS21]]</p>
+''white-space/pre-line'' and on <a>WebVTT Ruby Text Objects</a> and <a>WebVTT Ruby Base Text
```
zcorpan: Why was this added?

index.bs
```diff
@@ -4546,17 +4587,21 @@ corresponding <a lt="text track cue">cue</a>'s <a>WebVTT cue text alignment</a>:
...
+Objects</a> to ''white-space/nowrap''. [[!CSS21]]</p>
 
 <p>The 'font-style' property on <a lt="WebVTT Italic Object">WebVTT Italic Objects</a> must be set
 to ''font-style/italic''.</p>
 
 <p>The 'font-weight' property on <a lt="WebVTT Bold Object">WebVTT Bold Objects</a> must be set to
 ''font-weight/bold''.</p>
 
+<p>The 'font-size' property on <a>WebVTT Ruby Text Objects</a> must be set to ''font-size/50%''.</p>
```
zcorpan: Why was this added?

dwsinger:
On the implementation question, we have two implementations; the independent one would have to 'change' from ignoring rb because it's not specified in VTT into ignoring it because in the cases of ruby that VTT supports, it's not needed (a hard-to-detect change). The webkit one would have to make the simple change to not strip rb from text before passing it to the HTML engine (where it is implemented), so it would appear in the DOM etc.

PR #348, opened 2017-06-15
Add a note about the future of ruby

### LINKED ISSUE #264: incomplete ruby implementation [I18N-ISSUE-431]

zcorpan:
Fixes #264.
Closes #347.


silviapfeiffer:
I'm ok with this, though maybe we should mention the development of Ruby-CSS with a reference?

zcorpan:
I added something about that, please take a look.

silviapfeiffer:
Perfect, thanks. :-)

PR #356, opened 2017-08-17
Editorial: Clarify CSS escapes in an example

### LINKED ISSUE #352: What are "CSS character escape sequences"?
Re Example 6 https://www.w3.org/TR/2017/WD-webvtt1-20170808/#introduction-other-features

Would it be possible to link the text "CSS character escape sequences" to an explanation?

It took me a while to realise that `\31` related to the Unicode code point of `1`.

https://github.com/w3c/webvtt/blob/gh-pages/index.bs#L394

zcorpan:
Fixes #352.


silviapfeiffer:
Can we add a link to the CSS explanation fix mentioned in the bug? Looking good otherwise.

silviapfeiffer:
Sorry, I meant: a link to the doc.

zcorpan:
Ah yes, done.

PR #390, opened 2017-10-01
Move definition of chapter title text

### LINKED ISSUE #362: Wide Review Comment 2017:  Chapter location for defining WebVTT chapter title text
The term `WebVTT chapter title text` is defined in the section `4.6. Types of WebVTT files`. This should go in chapter `4.2 Types of WebVTT cue payload` (e.g. as `4.2.3`).

silviapfeiffer:
Move definition of chapter title text to a more appropriate location.
Closes #362

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1646,8 +1646,8 @@ characters (&amp;), and U+003E GREATER-THAN SIGN characters (>).</p>
 
 <h4 id=chapter-title-text>WebVTT chapter title text</h4>
 
-<p><dfn>WebVTT chapter title text</dfn> is <a>WebVTT cue text</a> that makes use only of zero or
-more of the following components, each optionally separated from the next by a <a>WebVTT line
+<p><dfn>WebVTT chapter title text</dfn> is <a>WebVTT cue text</a> that makes use of zero or more of
+the following components, each optionally separated from the next by a <a>WebVTT line
 terminator</a>:</p>
 
 <ul>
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
Andreas is ok with the patch, see https://github.com/w3c/webvtt/issues/362#issuecomment-336641648 - I'd like to land but need a review, @dwsinger

index.bs
```diff
@@ -1644,6 +1644,18 @@ than U+000A LINE FEED (LF) characters, U+000D CARRIAGE RETURN (CR) characters, U
 characters (&amp;), and U+003E GREATER-THAN SIGN characters (>).</p>
 
 
+<h4 id=chapter-title-text>WebVTT chapter title text</h4>
+
+<p><dfn>WebVTT chapter title text</dfn> is <a>WebVTT cue text</a> that makes use only of zero or
+more of the following components, each optionally separated from the next by a <a>WebVTT line
```
nigelmegitt: If the "only" applies to the following components, maybe rephrase as "makes use of zero or more of the following components, and no others, each ..."

silviapfeiffer: @nigelmegitt what is the alternative way of interpreting that? What's the source of confusion?

nigelmegitt: @silviapfeiffer it looks as though the "only" applies to the number of components, not the list from which that number could be selected. Saying "only of zero or more" looks odd, because it would be more natural to say "of zero or more", or "of any number". "Only" seems a bit of a sticky word, in that it feels like it is a qualifier that applies to the adjacent thing, rather than something later in the sentence. So if you say "of zero of more of the following components only, ..." then the "only" clearly applies to "the following components".

silviapfeiffer: I see. Maybe we could just remove the word "only"?

nigelmegitt: Yes, that would work.

PR #394, opened 2017-10-15
Add some explanatory text about the different types of VTT files.

### LINKED ISSUE #363: Wide Review Comment 2017: Distinction between cue payload types
In most cases it can not be said if cue payload is `WebVTT metadata text`, `WebVTT cue text` (in which case it would be of payload type `WebVTT cue text`) or `WebVTT chapter title text`. The only difference seems to be the allowance/disallowance of the AMPERSAND character and the LESS-THAN SIGN character. 

According to the file types definition in section 4.6 a WebVTT file using cue text is also a WebVTT file using metadata content and a WebVTT file using chapter title text. It seems that the distinction makes only sense if this is signaled externally.

silviapfeiffer:
Also explain that their interpretation is done by the environment,
i.e. the software that uses the file.

Closes #363

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -164,6 +164,12 @@ WEBVTT
 &lt;v Roger Bingham>You know I'm so excited my glasses are falling off here.
 </pre>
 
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
ping @tairt for review

PR #395, opened 2017-10-15
Add default color and background color names for applicable classes.

### LINKED ISSUE #365: Wide Review Comment 2017: Text color must be mandatory
User agents that render WebVTT must support text color. As the conformance class "User agents that don’t support CSS" is mentioned, this type of user agent would currently not support text color (see also https://github.com/w3c/webvtt/pull/279 ).

silviapfeiffer:
This brings in a capability to provide named colors and background colors
for captions and subtitles, matching what is being used by CEA708:
https://dvcs.w3.org/hg/text-tracks/raw-file/tip/608toVTT/608toVTT.html
I decided to add black and blue as per CEA708, but add a warning
to not use them on default background.

Closes #365


index.bs
```diff
@@ -4639,6 +4639,121 @@ settings:</p>
...
+   <td>''rgba(0,0,255,1)''</td>
+  </tr>
+  <tr>
+   <td><code>black</code></td>
+   <td>''rgba(0,0,0,1)''</td>
+  </tr>
+ </tbody>
+</table>
+
+<p class=note>The color for the class <code>green</code> is <em>not</em> the same as the CSS named
```
nigelmegitt: This issue has come up before and the TTWG resolved to retain the CSS definition of `"green"` in TTML but note that other standards' "green" is in fact equivalent to `"lime"` in CSS. This was in the context of TTML, but it makes sense to adopt the same approach here I think. I think it's confusing to have named colors that are other than those in CSS, although for TTML we did introduce some synonyms like "cyan" which is the same as "aqua".

index.bs
```diff
@@ -4639,6 +4639,121 @@ settings:</p>
...
+color ''green'', but rather matches ''lime''. This is because this color has traditionally been called ''green'' in captioning and subitling.</p>
+
+<p class=note>Do not use the classes <code>blue</code> and <code>black</code> on the default dark background, since they result in unreadable text.</p>
```
nigelmegitt: Maybe refer to WCAG guidance on color contrast as setting the requirements here, and in the context of captions on video, take account of the video's color too. There are plenty of other video/background/foreground color combinations that are not readable.

index.bs
silviapfeiffer: Curious to hear from others, since "green" in captioning has such an ingrained use. WebVTT cues are built as a captioning spec and CSS is just used for rendering so we could keep the captioning history. OTOH we could also use this to transform captioning practice to the more widely used CSS color names.

index.bs
silviapfeiffer: Good idea thanks.

hober:
As far as I can tell, this change is unnecessary at the VTT level. It belongs in the CEA608/708 mapping document. VTT already allows foreground and background color to be set, and arbitrary classes to be applied to cues. These are all the hooks the mapping document needs. It's the responsibility of the mapping document to define its own class names & default styling of them.

silviapfeiffer:
@hober the request was to have these classes supported in WebVTT so they interoperate between browsers and work for all WebVTT files everywhere, not just when you're trying to convert to 608/708. But I see your point.

nigelmegitt:
@hober if you mean that the change of definition of `green` to be different to the CSS definition is unnecessary in VTT then I agree.

silviapfeiffer:
To be clear: we're not changing what "green" means in CSS, we're just using the class "green" in WebVTT to mean the Captioning color green.

dwsinger:
I wonder if we can avoid the layering problem that hover points out by recommending that all VTT implementations support some suitable pre-defined styles, with a recommendation that those styles be the ones that define color defined in the style-sheet in https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html ?
(Implementations that support CSS could simply load that style-sheet; other implementations could recognize those style names and implement them.)

silviapfeiffer:
@dwsinger that's essentially the core of this PR except that we don't mention the word "style sheet". More importantly though is that we want to imply that there's a limit to what can be used in the STYLE section of WebVTT so non-browsers don't have to implement everything.

dwsinger:
Yes, but I am suggesting (a) not mentioning the colors directly here (b) making the 608/708 sets only the recommended ones (c) and using exactly those style names, rather than a new color property. So simply use <c.green> rather than a property, for example.

nigelmegitt:
Additional to your suggestion @dwsinger I am suggesting using the name `lime` rather than `green` for the style names and possibly adding a note for conversion from 608/708 that the binary value described as "green" in those specs needs to be mapped to the class name "lime". The motivation for doing this is to avoid downstream confusion caused by the class name not matching the CSS color value, or the named colors used anywhere else on the web.

dwsinger:
@nigelmegitt fine by me. @tairt would this approach work for you?

silviapfeiffer:
I think there is a misunderstanding of what @hober asked for. I will try to explain (correct me if I'm wrong).

@hober suggests we do colors this way:

```
STYLE
::cue(.white) { color: white; }
::cue(.green) { color: lime; }
::cue(.cyan) { color: cyan; }
::cue(.red) { color: red; }
::cue(.yellow) { color: yellow; }
::cue(.magenta) { color: magenta; }
::cue(.blue) { color: blue; }
::cue(.black) { color: black; }
::cue(.bg_white) { background: white; }
::cue(.bg_green) { background: lime; }
::cue(.bg_cyan) { background: cyan; }
::cue(.bg_red) { background: red; }
::cue(.bg_yellow) { background: yellow; }
::cue(.bg_magenta) { background: magenta; }
::cue(.bg_blue) { background: blue; }
::cue(.bg_black) { background: black; }
```

So, rather than pre-defining them as existing classes that can be used by a WebVTT caption author (which is what this patch does), we require every author who wants to use these to add them into the STYLE definition of their WebVTT file.

I believe it's @tairt 's desire to have these pre-definied so all authors use the same color class names. Are we in agreement that this is the desired outcome?

hober:
Oh, no, that wasn't what I was suggesting.

I think 608.css should define those styles, and the 608 mapping document should require conforming implementations that support CSS to have 608.css preloaded. (And it should say that implementations of the 608 spec that aren't based on CSS should do something equivalent.)

If the 608 spec did that, the VTT spec would need to do nothing at all.

hober:
I suppose it would be OK for the VTT spec to say something like "VTT processors SHOULD provide some mechanism for colors to be used in VTT documents. For example, VTT processors intending to comply with CEA-608 or 708 requirements could automatically provide the class names defined in *link to the mapping document*."

dwsinger:
@tairt would hober's suggestion be acceptable to you? essentially, we'd be recommending that all VTT UAs behave as if a suitable style-sheet were pre-loaded (and the only one out there is the one in the mapping document right now). Such a statement would also potentially trigger specs that use VTT to say something stronger (e.g. we could amend MPEG CMAF).

tairt:
The origin of the issue is the new conformance class "UA that do not support CSS". This new conformance class is only acceptable if VTT defines a way to support the basic captioning colors at least for text colors, independentely of CSS support. The issue was not about an interoperable color set and was not aiming at UA that DO support CSS. These UA's are of out of scope of the request because they are already meeting the requirements.

The current PR from Silvia works for me (including the mapping of the class name green to lime). I agree with @silviapfeiffer that the mapping need not to be the same as in CSS because the context is different. A compromise would be to leave out green completely and just use the class name "lime".

I can not see any advantage in refering to an VTT external mapping document. First the requirement for color support is not limited to 608 style captions (which are not used in many world regions, including Europe). Although accessibility requirements overlap between regions, this sets a wrong context. Second text color is a core feature which must be supported under all circumstances by all VTT User Agents. If VTT is a captioning spec this should out of question.

De-facto UA's without CSS (and hence with no color) support are already in the market. This causes accessibility issues. This is very serious because it affects not only these UA's. It is not un-common that content is produced for the lowest common technical demoninator. So there is the risk that color information is missing for captions for all plattforms/UAs.

silviapfeiffer:
Resolution at TPAC: we will add the style-class mappings to the appendix and require all UAs that implement WebVTT support to make these classes available to authors

silviapfeiffer:
Further resolution: call the lime color "lime" and add a note that 608/708 needs to convert to this

silviapfeiffer:
Resolution at TPAC: loading order is that of a UA stylesheet

silviapfeiffer:
I've applied all the changes requested at TPAC, except I made it a proper section because authors need to easily find this and so it can't be hidden behind the parsing and rendering in an appendix.

Please review patch.

silviapfeiffer:
@dwsinger I need a review on this patch

tairt:
I think the latest changes do not reflect exactly what we discussed. Now class names are reserved and they imply special semantics. As I understood @hober this is exactly what should be avoided. Although the result is the same the solution (as far as I understood) should be as follows:

* Define a stylesheet with cue pseudo elements and the class names for the colors as selectors.
* Add this stylesheet to a separate section of the spec.
* Mandate that all UA need to act as if the stylesheet has been pre-loaded.

silviapfeiffer:
@tait I don't understand the difference between what I specified and what you requested. The class names as specified are selectors. And I wrote that browsers should implement this as a user agent style sheet. And it's mandated because it's in the spec.

The only thing I didn't do is put it in an appendix because no author will find it there.

silviapfeiffer:
Comment from @hober very welcome!

tairt:
@hober Could you review the change as suggested by @silviapfeiffer? This would really help as I am not sure if the latest changes address your concerns.

tairt:
I think we should not merge before @hober confirms that this addresses his concerns.

silviapfeiffer:
Sure, no worries. I've been waiting for @hober to confirm. I'm just updating with the latest patches that landed.

hober:
Sorry for the delay (Thanksgiving break, then recovery from a medical procedure, have kept me away.)

@tairt is correct that my issue was with reserving specific class names and imbuing them with semantics. And @silviapfeiffer is right that what we agreed to at TPAC is doing exactly that anyway.

I think this pull request implements what was agreed to at TPAC.

silviapfeiffer:
@hober thanks for clarifying!

tairt:
For me the PR and solution is fine. But I may not fully understand the issue that was raised by @hober 

> @tairt is correct that my issue was with reserving specific class names and imbuing them with semantics. 

Isn't [Section 5.1](https://w3c.github.io/webvtt/#default-text-color) doing exactly this?

One other question: if the class names listed in 5.1 are set by an author with a ::cue pseudo element they get overwritten, correct? If this is the case...is this clear enough in the spec?

silviapfeiffer:
Yes, we are now reserving specific class names, but I think @hober is now ok with it.

Yes, if you overwrite a default class's color through ::cue(class) specifications. I can add a note about that.

silviapfeiffer:
@tairt please check #418

PR #399, opened 2017-10-28
Clarify line cue setting specification.

### LINKED ISSUE #382: Wide Review Comment 2017: line cue setting direction
Copy/paste from https://lists.w3.org/Archives/Public/public-tt/2017Sep/0080.html - raising as an issue for tracking/disposition purposes.

(Context is section 4.4 "WebVTT cue settings" -> the "WebVTT line cue setting")

looking at that line cue setting, it "configures the offset of the cue box from the video viewport's edge in the direction opposite to the writing direction". I think that means "orthogonal to the writing direction" but it could be interpreted as the negative of the writing direction, i.e. If the writing direction is top-to-bottom then does this mean left-to-right or bottom-to-top? 

Then the WebVTT position cue setting "configures the indent position of the cue box in the direction orthogonal to the WebVTT line cue setting". Why doesn't it simply say "in the writing direction"?

silviapfeiffer:
Closes #382

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1863,15 +1863,16 @@ given:</p>
 </ol>
 
 <p class="note">A <a>WebVTT line cue setting</a> configures the offset of the cue box from the video
-viewport's edge in the direction opposite to the <a lt="WebVTT cue writing direction">writing
-direction</a>. For horizontal cues, this is the vertical offset from the top of the video viewport.
-The offset is for the <a lt="WebVTT cue line start alignment">start</a>, <a lt="WebVTT cue line
-center alignment">center</a>, or <a lt="WebVTT cue line end alignment">end</a> of the cue box,
-depending on the <a>WebVTT cue line alignment</a> value - <a lt="WebVTT cue line start
-alignment">start</a> by default. The offset can be given either as a percentage of the video
-dimension or as a line number. Line numbers are based on the size of the first line of the cue.
-Positive line numbers count from the start of the video viewport (the first line is numbered 0),
-negative line numbers from the end of the viewport (the last line is numbered &#x2212;1).</p>
+viewport's edge in the direction orthogonal to the <a lt="WebVTT cue writing direction">writing
+direction</a>. For horizontal cues, this is the vertical offset from the top of the video viewport,
+for vertical cues, it's the horizontal offset. The offset is for the <a lt="WebVTT cue line start
+alignment">start</a>, <a lt="WebVTT cue line center alignment">center</a>, or <a lt="WebVTT cue line
+end alignment">end</a> of the cue box, depending on the <a>WebVTT cue line alignment</a> value - <a
+lt="WebVTT cue line start alignment">start</a> by default. The offset can be given either as a
+percentage of the video dimension or as a line number. Line numbers are based on the size of the
+first line of the cue. Positive line numbers count from the start of the video viewport (the first
+line is numbered 0), negative line numbers from the end of the viewport (the last line is numbered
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
@dwsinger review required

index.bs
```diff
@@ -1873,15 +1873,16 @@ given:</p>
...
-dimension or as a line number. Line numbers are based on the size of the first line of the cue.
-Positive line numbers count from the start of the video viewport (the first line is numbered 0),
-negative line numbers from the end of the viewport (the last line is numbered &#x2212;1).</p>
+viewport's edge in the direction orthogonal to the <a lt="WebVTT cue writing direction">writing
+direction</a>. For horizontal cues, this is the vertical offset from the top of the video viewport,
+for vertical cues, it's the horizontal offset. The offset is for the <a lt="WebVTT cue line start
+alignment">start</a>, <a lt="WebVTT cue line center alignment">center</a>, or <a lt="WebVTT cue line
+end alignment">end</a> of the cue box, depending on the <a>WebVTT cue line alignment</a> value - <a
+lt="WebVTT cue line start alignment">start</a> by default. The offset can be given either as a
+percentage of the video dimension or as a line number. Line numbers are based on the size of the
```
nigelmegitt: This is the size in the block progression direction, i.e. the vertical height for horizontal cues, right?

index.bs
```diff
@@ -1873,15 +1873,16 @@ given:</p>
...
+first line of the cue. Positive line numbers count from the start of the video viewport (the first
+line is numbered 0), negative line numbers from the end of the viewport (the last line is numbered
```
nigelmegitt: Is is true that start == (top or left) and end == (bottom or right)? Or are those terms writing mode dependent, i.e. end could be the left in rtl?

index.bs
silviapfeiffer: yes.

index.bs
silviapfeiffer: I'll quote from the spec, which answers your question:

> A line alignment
> An alignment for the cue box’s line, one of:

> Start alignment
> The cue box’s top side (for horizontal cues), left side (for vertical growing right), or right side (for vertical growing left) is aligned at the line.

> Center alignment
> The cue box is centered at the line.

> End alignment
> The cue box’s bottom side (for horizontal cues), right side (for vertical growing right), or left side (for vertical growing left) is aligned at the line.

>By default, the line alignment is set to start.

The line alignment is separate from the text alignment — right-to-left vs. left-to-right cue text does not affect the line alignment.

index.bs
silviapfeiffer: Yes.

index.bs
silviapfeiffer: Yes, writing mode dependent as is written in the spec.

index.bs
nigelmegitt: Thanks, I think the bit I was focusing on here was "percentage of the video dimension" which seems vague, since it's a two dimensional thing, and in general not square. Suggestion would be to qualify it by saying something like "percentage of the relevant writing-mode dependent video dimension"

index.bs
nigelmegitt: Thanks, I thought so. I'm not sure why I was asking now, probably I was worried about the meaning of "start" being clear, but it is well defined. This looks fine to me now.

index.bs
silviapfeiffer: OK

silviapfeiffer: @nigelmegitt please see PR #420

PR #407, opened 2017-11-06
Add code of conduct as proposed by plh

nigelmegitt:
No ticket - PLH pointed out that we need code of conduct files on all repos, so this PR just adds one.


silviapfeiffer:
👍

PR #408, opened 2017-11-10
Turn long alt text into longdesc.

### LINKED ISSUE #386: Change alt text for region diagram
Feedback from the Accessible Platform Architectures (APA) Working Group about the new WebVTT spec:

You have an alt= that is far too long, in APA's view. We would suggest
using Figure/Details (or even a longdesc) instead. The code involved is
the following:

   <div class="note" role="note">
    <p>The following diagram illustrates how anchoring of a region to a video viewport works. The black
 cross is the anchor, orange explains the anchor's offset within the region and green the anchor's
 offset within the viewport. Think of it as sticking a pin through a note onto a board:</p>
    <p>Within the video viewport, there is a WebVTT region.
 Inside the region, there is an anchor point marked with a black cross. The vertical and horizontal
 distance from the video viewport's edges to the anchor is marked with green arrows, representing
 the region viewport anchor X and Y offsets. The vertical and horizontal distance from the region's
 edges to the anchor is marked with orange arrows, representing the region anchor X and Y offsets.
 The size of the region is represented by the region width for the horizontal axis, and region lines
 for the vertical axis.</p>
   </div>

The problem, of course, with long alt= is that the user who relies on
alt= is forced to digest the complete text each and every time they
source the page. With longish alt= that becomes quite annoying.

That's why we have Figure/Details (and longdesc), as you may well
recall. We want the textual description. We just don't want to enforce
it without the user asking to see the details.

silviapfeiffer:
This closes #386

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1207,13 +1207,15 @@ area for <a lt="WebVTT cue">WebVTT cues</a>.</p>
  <p>The following diagram illustrates how anchoring of a region to a video viewport works. The black
  cross is the anchor, orange explains the anchor's offset within the region and green the anchor's
  offset within the viewport. Think of it as sticking a pin through a note onto a board:</p>
- <p><img src="webvtt-region-diagram.png" alt="Within the video viewport, there is a WebVTT region.
+ <p><img src="webvtt-region-diagram.png" alt="visual explanation of WebVTT regions"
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
@dwsinger request for review

Issue #383, opened 2017-09-27
Wide Review Comment 2017: positioning model


nigelmegitt:
Copy/paste from https://lists.w3.org/Archives/Public/public-tt/2017Sep/0080.html - raising as an issue for tracking/disposition purposes.

It would be easier for me to understand and follow if there were a clearer box model, in which there is a region box, a cue box within that region, and then a set of cue lines within that cue box. Then the positioning units of, say, the cue box, can be in terms of lines, relative to the region edges, and the overall rendering area (the default region?) can be positioned relative to the video or the video viewport in a consistent manner.

silviapfeiffer:
We can add some descriptive text to section 3 where the cue and region are introduced. Essentially it's as follows:

The box model of WebVTT consists of three elements: the video viewport, cues and regions. The video viewport is the rendering area into which cues and regions are rendered. Cues are boxes consisting of a set of cue lines. Regions are subareas of the view viewport that are used to group cues together. Cues are positioned either relative to the video viewport or relative to a region.

Would that be sufficient?

dwsinger:
The box model of WebVTT consists of three elements: the video viewport, cues, and regions. The video viewport is the rendering area into which cues and regions are rendered. Cues are boxes consisting of a set of cue lines. Regions are subareas of the video viewport that are used to group cues together. Cues are positioned either relative to the video viewport or inside a region. Regions are positioned in the video viewport.

silviapfeiffer:
Resolution at TPAC: Silvia to work with Nigel to resolve the wording - this is the right approach

dwsinger:
(editors to finesse with the commenter. maybe put the text into encapsulation order, for clarity...)

silviapfeiffer:
explain how the setting work on cues and regions and which are applied as group and single

silviapfeiffer:
@nigelmegitt please review the PR

nigelmegitt:
@silviapfeiffer done

dwsinger:
added @nigelmegitt commenter-agreed and closing

silviapfeiffer:
https://github.com/w3c/webvtt/pull/409 is the Telebart PR and it hasn't landed yet...

nigelmegitt:
Also I haven't completed the (re-)review so this was too quick.

silviapfeiffer:
#409 has landed

PR #409, opened 2017-11-11
Add some introductory paragraphs to the Data model section.

silviapfeiffer:
This is to explain the box model and how settings work on cues and regions.

closes issue #383

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,3 +22,7 @@ To generate a WD snapshot, run e.g.:
     $ ./snapshot.sh WD 2016-01-01 2015-12-08
 
 Also see https://github.com/w3c/webvtt/commit/754f13e3cf03d6036c3e4628c6920d17b412f778 for manual fixup of the generated output.
+
+To format the index.bs file, run:
+
+  $ ./format.py index.bs
--- a/index.bs
+++ b/index.bs
@@ -669,11 +669,11 @@ boxes consisting of a set of cue lines. Regions are subareas of the video viewpo
 group cues together. Cues are positioned either inside the video viewport directly or inside a
 region, which is positioned inside the video viewport.</p>
 
-<p class=note>A set of cue settings position cues inside the video viewport. A set of region
-settings position regions inside the video viewport. Cues that are inside regions can only use a
-limited set of their cue settings. Specifically, if the cue has a "vertical", "line" or "size"
-setting, the cue drops out of the region. Otherwise, the cue's width is calculated to be relative to
-the region width rather than the viewport. </p>
+<p class=note>The position of a cue inside the video viewport is defined by a set of cue settings.
+The position of a region inside the video viewport is defined through a set of region settings. Cues
+that are inside regions can only use a limited set of their cue settings. Specifically, if the cue
+has a "vertical", "line" or "size" setting, the cue drops out of the region. Otherwise, the cue's
+width is calculated to be relative to the region width rather than the viewport. </p>
 
 
 <h3 id=cues>WebVTT cues</h3>
--- a/index.bs
+++ b/index.bs
@@ -1187,6 +1187,10 @@ area for <a lt="WebVTT cue">WebVTT cues</a>.</p>
  <dd>
   <p>A number giving the number of lines of the box within which the text of each line of the
   containing cues is to be rendered. Defaults to 3.</p>
+
+  <p class="note">Since a WebVTT region defines a fixed rendering area, a cue that has more lines
+  than the region allows will be clipped. For scrolling regions, the clipping happens at the top,
+  for non-scrolling regions it happens at the bottom.</p>
  </dd>
 
  <dt><dfn lt="WebVTT region anchor">A region anchor point</dfn></dt>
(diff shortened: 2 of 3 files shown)


silviapfeiffer:
@dwsinger please review

index.bs
```diff
@@ -663,6 +663,18 @@ precomposed character).</p>
 <!-- Add some content here about cues and serialisation format in general -->
 <!-- Describe metadata, caption/subtitle, chapter & description cues -->
 
+<p class=note>The box model of WebVTT consists of three key elements: the video viewport, cues, and
+regions. The video viewport is the rendering area into which cues and regions are rendered. Cues are
+boxes consisting of a set of cue lines. Regions are subareas of the video viewport that are used to
+group cues together. Cues are positioned either inside the video viewport directly or inside a
+region, which is positioned inside the video viewport.</p>
+
+<p class=note>A set of cue settings position cues inside the video viewport. A set of region
```
nigelmegitt: > A set of cue settings position cues inside the video viewport.

is really hard to parse. How about:

_The positions of cues inside the video viewport are defined by a set of cue settings._

index.bs
```diff
@@ -663,6 +663,18 @@ precomposed character).</p>
...
+settings position regions inside the video viewport. Cues that are inside regions can only use a
+limited set of their cue settings. Specifically, if the cue has a "vertical", "line" or "size"
```
nigelmegitt: Likewise: 

_The positions of regions inside the video viewport are defined by a set of region settings._

index.bs
```diff
@@ -663,6 +663,18 @@ precomposed character).</p>
...
+setting, the cue drops out of the region. Otherwise, the cue's width is calculated to be relative to
+the region width rather than the viewport. </p>
```
nigelmegitt: I'd like to see an extra sentence in here clarifying if regions act as clipping area for the cues within them or if they are only positioning constructs.

index.bs
silviapfeiffer: ok, thanks.

index.bs
silviapfeiffer: ok.

index.bs
silviapfeiffer: "Clipping" is not the right word: if a cue text is longer than the region width allows, it will be broken into another line - that's what CSS does for us.

The definition of region width already states that a cue is to be rendered within the region's width: https://www.w3.org/TR/webvtt1/#webvtt-region-width

> A width
> A number giving the width of the box within which the text of each line of the containing cues is to be rendered, to be interpreted as a percentage of the video width. Defaults to 100.

You'll notice that if the cue has an explicit "size" setting, it drops out of the region. A cue's default size is 100%. For regions this is relative to the region width. Therefore, the cue's width is the same as the region's width.

How would you like me to say this?

silviapfeiffer:
Please check the last comment - first two comments have been addressed.

index.bs
nigelmegitt: Gosh, right, there are lots of things to unpack here.

Firstly, in terms of clipping behaviour I think you're saying that there is a combination of wrapping and clipping, so that the clipping only ever occurs in the block progression direction, by means of losing whole lines. The consequence is that no cue text is ever displayed outside the region that the cue is in. However if the text size is such that the whole cue does not fit within the region, there may be some lines of text that are _never visible_. The general rule is that lines that do not fit in their cue's region are not displayed. Those points are worth stating.

Secondly, I can't see in the parsing and processing models why cues with a "size" setting drop out of the region. If they drop out, where do they go? Are they dropped on the floor, or displayed somewhere else outside the region?

The same is true for the "vertical" and "line" settings: in all three cases I could not find algorithmic steps switched on the presence or absence of those settings that modify the position to be either region-based or viewport-based. Please could you point me to where that happens?

index.bs
```diff
@@ -669,11 +669,11 @@ boxes consisting of a set of cue lines. Regions are subareas of the video viewpo
 group cues together. Cues are positioned either inside the video viewport directly or inside a
 region, which is positioned inside the video viewport.</p>
 
-<p class=note>A set of cue settings position cues inside the video viewport. A set of region
-settings position regions inside the video viewport. Cues that are inside regions can only use a
-limited set of their cue settings. Specifically, if the cue has a "vertical", "line" or "size"
-setting, the cue drops out of the region. Otherwise, the cue's width is calculated to be relative to
-the region width rather than the viewport. </p>
+<p class=note>The position of a cue inside the video viewport is defined by a set of cue settings.
+The position of a region inside the video viewport is defined through a set of region settings. Cues
```
nigelmegitt: Probably we should be consistent about the use of "by" or "through". I'd go for "by" in both places.

silviapfeiffer: OK :-)

index.bs
silviapfeiffer: "Dropping out of a region" merely means that they are not rendered within the region, but rendered in their own right with the settings given.

Looks like there are some things missing in the parsing algorithm. I'll add some text. Thanks for noticing.

silviapfeiffer:
@nigelmegitt ready for re-review

index.bs
nigelmegitt: You're welcome. By the way, this is _exactly_ the kind of thing that's in my mind when I grump about algorithmic specifications and the ability for the WG to test that they do what they're supposed to do.

README.md
```diff
@@ -22,3 +22,7 @@ To generate a WD snapshot, run e.g.:
     $ ./snapshot.sh WD 2016-01-01 2015-12-08
 
 Also see https://github.com/w3c/webvtt/commit/754f13e3cf03d6036c3e4628c6920d17b412f778 for manual fixup of the generated output.
+
+To format the index.bs file, run:
+
+  $ ./format.py index.bs
```
nigelmegitt: This seems like it has nothing to do with #383, but whatever, no big deal.

silviapfeiffer: Yeah, it's just for future editors - had to come in somewhere. ;-)

Issue #391, opened 2017-10-04
VLC feedback: explicitly list CSS features supported in WebVTT


silviapfeiffer:
Particularly for user agents that do not support CSS, it's important to list all the CSS features that are supported in CSS, particularly in the STYLE element.

dwsinger:
Not sure I completely understand. There are UAs that don't do CSS; all you get is the basic nuilt-in styling support (which needs to include color, and currently has stuff like <i>). There are UAs that do support CSS, and I suppose we could list the CSS that they need to support (2.1?), but should we?

silviapfeiffer:
There are some css selectors that are allowed to be applied to cues and spans. We should give non CSS user agents a chance to implement those for use inside the STYLE tag. But we need to list them more explicitly in the format section and take out of the rendering section.

dwsinger:
We could say that a non-CSS user-agent may implement selected CSS features, or even should implement simple ones that we mention in the document and list, sure.

silviapfeiffer:
Yes, indeed 😀

PR #389, opened 2017-10-01
Add a "User agents that do not support CSS" conformance class.

silviapfeiffer:
This actually introduces a definition for "WebVTT cue setting" and
clarifies how they provide information to position and align the cue
box and cue text.

Closes #361

PR #410, opened 2017-11-11
Add a paragraph about the reasons why limiting styles.

### LINKED ISSUE #373: Wide Review Comment 2017: future styling options mechanism
Copy/paste from https://lists.w3.org/Archives/Public/public-tt/2017Sep/0080.html - raising as an issue for tracking/disposition purposes.

The list of ::cue CSS properties in §7.2.1 is a useful minimal set, especially for a non-CSS-based implementation however the requirement to ignore any properties not in the list seems overly restrictive for a CSS based implementation. What is the mechanism for adding further styling options in the future, given that WebVTT files do not indicate any specification version?

silviapfeiffer:
closes #373

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -756,7 +756,7 @@ requirements.</p>
 
  <dt><dfn>User agents that do not support a full HTML CSS engine</dfn></dt>
  <dd><p>All processing requirements in this specification apply, including the color classes defined
- in [[#default-classes]]. However, the user agent will need to interpret the CSS related features in
+ in [[#default-classes]]. However, the user agent will need to apply the CSS related features in
  [[#parsing]], [[#rendering]] and [[#css-extensions]] in such a way that the rendered results are
  equivalent to what a full CSS supporting renderer produces.</p></dd>
 
@@ -764,8 +764,8 @@ requirements.</p>
  <dd><p>All processing requirements in this specification apply. However, only a limited set of CSS
  styles is allowed because <a>user agents that do not support a full HTML CSS engine</a> will need
  to implement CSS functionality equivalents. User agents that support a full CSS engine must
- therefore limited the CSS styles they interpret for WebVTT so as to enable identical rendering
- without bleeding in extra CSS styles that are beyond the WebVTT specification.</p></dd>
+ therefore limit the CSS styles they apply for WebVTT so as to enable identical rendering without
+ bleeding in extra CSS styles that are beyond the WebVTT specification.</p></dd>
 
  <dt>Conformance checkers</dt>
  <dd><p>Conformance checkers must verify that a <a>WebVTT file</a> conforms to the applicable
--- a/index.bs
+++ b/index.bs
@@ -307,9 +307,9 @@ target WebVTT cues and regions in the video using the ''::cue'', ''::cue()'', ''
 ''::cue-region()'' pseudo-elements.</p>
 
 <p>Only a limited set of CSS styles is allowed in WebVTT because user agents that do not support CSS
-will need to implement CSS functionality equivalents. Browsers therefore also have to be limited in
-what CSS styles they support for WebVTT so as to enable identical rendering in browsers and
-non-browsers.</p>
+will need to implement CSS functionality equivalents. User agents that support a full CSS engine
+must limited the CSS styles they interpret for WebVTT so as to enable identical rendering without
+bleeding in extra CSS styles that are beyond the WebVTT specification.</p>
 
 <div class="example">
 
--- a/index.bs
+++ b/index.bs
@@ -281,11 +281,6 @@ one line except when a line break is definitely necessary.</p>
 target WebVTT cues and regions in the video using the ''::cue'', ''::cue()'', ''::cue-region'' and
 ''::cue-region()'' pseudo-elements.</p>
 
-<p>Only a limited set of CSS styles is allowed in WebVTT because user agents that do not support CSS
-will need to implement CSS functionality equivalents. User agents that support a full CSS engine
-must limited the CSS styles they interpret for WebVTT so as to enable identical rendering without
-bleeding in extra CSS styles that are beyond the WebVTT specification.</p>
-
 <div class="example">
 
  <p>In this example, an HTML page has a CSS style sheet in a <a element>style</a> element that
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
@dwsinger need a review

index.bs
```diff
@@ -302,6 +302,11 @@ one line except when a line break is definitely necessary.</p>
 target WebVTT cues and regions in the video using the ''::cue'', ''::cue()'', ''::cue-region'' and
 ''::cue-region()'' pseudo-elements.</p>
 
+<p>Only a limited set of CSS styles is allowed in WebVTT because user agents that do not support CSS
+will need to implement CSS functionality equivalents. Browsers therefore also have to be limited in
+what CSS styles they support for WebVTT so as to enable identical rendering in browsers and
+non-browsers.</p>
```
nigelmegitt: I am not clear that this is a genuine requirement, that rendering is identical in browsers and non-browsers. If I recall correctly WebVTT has no requirements document. Pointer to a WG resolution that this is important?

dwsinger: Actually this sentence is subtly wrong; user agents that don't support CSS (a conformance class as Andreas notes), well, do not support CSS.  UAs that DO support CSS *can* use a full CSS engine but can also implement this restricted subset of CSS.

silviapfeiffer: OK, I'll reformulate.

silviapfeiffer:
review again :-)

index.bs
```diff
@@ -307,9 +307,9 @@ target WebVTT cues and regions in the video using the ''::cue'', ''::cue()'', ''
 ''::cue-region()'' pseudo-elements.</p>
 
 <p>Only a limited set of CSS styles is allowed in WebVTT because user agents that do not support CSS
-will need to implement CSS functionality equivalents. Browsers therefore also have to be limited in
-what CSS styles they support for WebVTT so as to enable identical rendering in browsers and
-non-browsers.</p>
+will need to implement CSS functionality equivalents. User agents that support a full CSS engine
+must limited the CSS styles they interpret for WebVTT so as to enable identical rendering without
```
nigelmegitt: It's not clear what the "identical rendering" is referring to here - identical to what?

dwsinger:
I think this is still wrong. User agents that don't support CSS don't support CSS, that's why we had to do the colour work.  This is about UAs that DO support CSS but DON'T have a full engine; they are only required to support this subset of CSS. Please stay silent on what happens if you have a full engine; I think requiring those implementations to be sure to limit themselves to that subset is adding unneeded pain.

index.bs
silviapfeiffer: Identical to what is rendered when you follow the spec and have CSS functionality.

silviapfeiffer:
@dwsinger No, this paragraph is not about UAs that support CSS partially. I don't think such a class exists. It's about the difference between UAs that support CSS and those that don't support CSS (like VLC). It says that those that support CSS (like Web browsers) must not allow the full capability of their CSS support in WebVTT and **this is something I am not going to budge on**. It is extremely important that we disallow full engines to allow all their functionality in WebVTT, because if we do, then everyone has to implement the full CSS engine for WebVTT.

Remember the times when Internet Explorer was the dominant Web browser and they implemented functionality that was not in the HTML&CSS standards? What did that do to the other browsers? This paragraph is about avoiding that situation, so we get compatible implementations.

dwsinger:
OK, there are two problems here, and we should disentangle them. (a) your text talks about non-CSS UAs that only implement a subset. Those are NOT non-CSS UAs, non-CSS UAs don't implement any CSS. (b) then there is the question of what CSS you are allowed to use and the UA is required to support, and whether the UA has to ignore anything else. I'm OK with a parse-time discard on (b). So, the text needs to talk about CSS-supporting UAs only supporting the CSS we document.

silviapfeiffer:
@dwsinger re (a) Read that sentence properly please. It says:

"Only a limited set of CSS styles is allowed in WebVTT because user agents that do not support CSS
will need to implement CSS functionality equivalents."

this is referring to:
1/ the WebVTT spec specifying a limited set of CSS styles
2/ UAs that don't support CSS
3/ CSS functionality **equivalents**

It does not say that these UAs implement a subpart of CSS. It says they have to implement functionality **equivalents** of the limited CSS subset defined in WebVTT.

ad (c) The second sentence says exactly what you are suggesting.

Could you please provide proposed wording changes that make it clearer for you? I really don't know how else to say it more clearly.

silviapfeiffer:
OK, sounds to me like we need to define our conformance classes better: one that has no CSS support, one that supports the CSS parts that WebVTT uses (see #391) and UAs with a full CSS engine.

silviapfeiffer:
Is also relate to #389

dwsinger:
possibly, but I think you argued that the reduced CSS and full CSS engines had to give the same result, so probably only two

silviapfeiffer:
Not quite. "WebVTT CSS" supporting UAs behave differently, since they don't follow the rendering section to the word. So, it needs an out from some of the spec. I think we need three levels.

1/ No CSS: non-CSS UAs that ignore all CSS

2/ WebVTT CSS: non-CSS UAs that implement functional equivalents to the WebVTT CSS features and therefore are allowed to not follow the text of the specification in the rendering section

3/ Full CSS: CSS-supporting UAs (mostly browsers) that implement the WebVTT CSS functionality as specified

silviapfeiffer:
Please review #389 . If that one's ok, then we can move forward to reformulate the paragraph here and base it on those three conformance classes. (Incidentally, the third one is not explicitly formulated in #389 because it's implied by the spec.)

silviapfeiffer:
Since #389 landed, I have now updated this PR. I decided to move the paragraph out of an example section into a normative space and it made sense to explicitly define the "Full CSS engine" conformance class for this. Please re-review.

index.bs
```diff
@@ -765,6 +760,13 @@ requirements.</p>
  [[#parsing]], [[#rendering]] and [[#css-extensions]] in such a way that the rendered results are
  equivalent to what a full CSS supporting renderer produces.</p></dd>
 
+ <dt><dfn>User agents that support a full HTML CSS engine</dfn></dt>
+ <dd><p>All processing requirements in this specification apply. However, only a limited set of CSS
+ styles is allowed because <a>user agents that do not support a full HTML CSS engine</a> will need
+ to implement CSS functionality equivalents. User agents that support a full CSS engine must
+ therefore limited the CSS styles they interpret for WebVTT so as to enable identical rendering
```
nigelmegitt: typo: s/limited/limit

Is "interpret" the right term here, addressing your concerns @dwsinger ? Seems to me that "apply" would be better.

dwsinger: yes, s/limited/limit/ and "limit the CSS they apply" might be better

silviapfeiffer: ok

silviapfeiffer:
Please re-review

dwsinger:
I'm (still) OK

PR #419, opened 2017-12-08
Timeranges

### LINKED ISSUE #415: Unclear definition of time range
The data model for WebVTT cues omits the begin and end times, because they are defined as extensions of [text track cues](https://www.w3.org/TR/html51/semantics-embedded-content.html#cue), whose definition is:

> **A start time**
>    The time, in seconds and fractions of a second, that describes the beginning of the range of the media data to which the cue applies. 
>
> **An end time**
>    The time, in seconds and fractions of a second, that describes the end of the range of the media data to which the cue applies. 

It is unclear in any of the definitions of the begin and end times are included in the range or excluded. They should be aligned with the TTML and SMIL definitions where the begin time is included in the range and the end time is excluded, so that if there are two adjacent cues where the first cue's end timestamp is equal to the second cue's begin timestamp then the cues' presentation does not overlap.

silviapfeiffer:
Closes #415

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1795,8 +1795,15 @@ replaced.</p>
 
 <h3 id=cue-settings>WebVTT cue settings</h3>
 
-<p>A <a>WebVTT cue settings list</a> consists of zero or more of the following settings. Each
-setting must not be included more than once per <a>WebVTT cue settings list</a>.</p>
+<p>A <a>WebVTT cue setting</a> is part of a <a>WebVTT cue settings list</a> and provides
+configuration options regarding the position and alignment of the cue box and the cue text
+within.</p>
+
+<p class="note">For example, a set of WebVTT cue settings may allow a cue box to be aligned to the
+left or positioned at the top right with the cue text within center aligned.</p>
+
+<p>The current available <a>WebVTT cue settings</a> that may appear in a <a>WebVTT cue settings
+list</a> are:</p>
 
 <ul class="brief">
  <li>A <a>WebVTT vertical text cue setting</a>.</li>
@@ -1807,9 +1814,8 @@ setting must not be included more than once per <a>WebVTT cue settings list</a>.
  <li>A <a>WebVTT region cue setting</a>.</li>
 </ul>
 
-<p class="note">A <a>WebVTT cue settings list</a> gives configuration options regarding the position
-and alignment of the cue box and the cue text within. For example, it allows a cue box to be aligned
-to the left or positioned at the top right with the cue text within center aligned.</p>
+<p>Each of these setting must not be included more than once per <a>WebVTT cue settings
+list</a>.</p>
 
 <p>A <dfn>WebVTT vertical text cue setting</dfn> is a <a>WebVTT cue setting</a> that consists of the
 following components, in the order given:</p>
--- a/index.bs
+++ b/index.bs
@@ -48,7 +48,9 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: rules for updating the text track rendering
             text: text track cue active flag
             text: text track cue text
+            text: text track cue data
             text: text track cue display state
+            text: time marches on
             text: current playback position
             text: text track cue identifier
             text: text track cue pause-on-exit flag
@@ -666,8 +668,14 @@ precomposed character).</p>
 
 <h3 id=cues>WebVTT cues</h3>
 
-<p>A <dfn>WebVTT cue</dfn> is a <a>text track cue</a> that additionally consist of the following:
-[[!HTML]]</p>
+<p>A <dfn>WebVTT cue</dfn> is a <a>text track cue</a> as defined in the HTML specification [[!HTML]]
+with <a lt="text track cue data">additional format-specific data</a>.</p>
+
+<p class=note>Importantly, this provides start and end time attributes to WebVTT cues. They are
+interpreted via the <a>time marches on</a> algorithm of the HTML specification to include the start
+time in the range and exclude the end time.</p>
+
+<p>The additional format-specific data for <a>WebVTT cues</a>is the following:</p>
 
 <dl>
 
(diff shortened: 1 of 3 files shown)


silviapfeiffer:
@nigelmegitt added as requested, please re-review

index.bs
```diff
@@ -3921,20 +3922,19 @@ follows:</p>
 
 <h2 id=rendering>Rendering</h2>
 
-<p class="note">This section describes in some detail how to visually render WebVTT cues in a user
-agent. The processing model is quite tightly linked to media elements in HTML. When supporting
-WebVTT in media players that don't support CSS, equivalent visual rendering will need to be
-implemented.</p>
+<p>This section describes how to visually render WebVTT cues in a user agent. The processing model
+is tightly linked to media elements in HTML and follows the handling of cues when <a>playing the
+media resource</a>. When supporting WebVTT in media players that don't support CSS, equivalent
```
nigelmegitt: "don't support CSS" now needs to be expanded to "don't support CSS and HTML5" I think

dwsinger: HTML5 is being used to define semantics of cue times; do I need to "support HTML5" to get that?

silviapfeiffer: @nigelmegitt how about "don't support HTML nor CSS" - I just want to make sure players that support CSS but not HTML are covered.

silviapfeiffer:
Please re-review

index.bs
nigelmegitt: Sorry @dwsinger I didn't spot your response a week ago. I think that you do not need to support HTML5 to match the semantics of cue times, in theory, though you might in practice!

nigelmegitt: @silviapfeiffer the correct wording would be "that support neither HTML nor CSS" I think, or alternatively "that do not support HTML and do not support CSS" to be super clear.

nigelmegitt: @silviapfeiffer sorry for the confusion, but I think we would need to clarify what implementations that support HTML but not CSS, or CSS but not HTML would do with my proposed wording at https://github.com/w3c/webvtt/pull/419#discussion_r170596448 and also in the more recent update. I suspect the better option is to say "that do not support both HTML and CSS".

dwsinger:
"Support neither CSS nor HTML" would be normal english.

nigelmegitt:
> "Support neither CSS nor HTML" would be normal english.

@dwsinger I had to think about the logic of what this statement is saying too. I believe the sentence applies to all processors that do not support both CSS and HTML, since if either one or both of those is not (fully) supported then the implementation is still expected to behave equivalently.

silviapfeiffer:
I think I'll go back to "don't support CSS" because it doesn't matter whether it supports HTML or not - this is about the rendering. "neither" makes it sound like if you support CSS, but not HTML, it doesn't apply to you, which isn't the case.

nigelmegitt:
@silviapfeiffer the Rendering section includes the processing model, which is dependent on HTML time marches on, so it is appropriate to refer to HTML support as well as CSS support here.

silviapfeiffer:
@nigelmegitt yes, indeed. That's why the first sentence addresses the lack of HTML: The processing model is tightly linked to media elements in HTML and follows the handling of cues when <a>playing the media resource</a>. The second sentence is just about the visual rendering, which is about CSS. But I can more explicitly refer to the conformance classes "User agents that do not support CSS"  and "User agents that do not support a full HTML CSS engine" to clarify this.

nigelmegitt:
> But I can more explicitly refer to the conformance classes "User agents that do not support CSS" and "User agents that do not support a full HTML CSS engine" to clarify this.

@silviapfeiffer sounds good to me. Just one more thought: are HTML and CSS completely separable in this case, so that there may be an HTML engine that does not support CSS, or is that so niche that it is not worth considering?

tairt:
@silviapfeiffer wrote

> I think I'll go back to "don't support CSS" because it doesn't matter whether it supports HTML or not - this is about the rendering. 

This works for me.

@silviapfeiffer wrote
> But I can more explicitly refer to the conformance classes "User agents that do not support CSS" and "User agents that do not support a full HTML CSS engine" to clarify this.

This work for me too.

As the second edit would also work for @nigelmegitt: @silviapfeiffer  could you do the edit so this PR can  be closed.

silviapfeiffer:
@nigelmegitt yes, there are HTML renderers that do not support a full CSS engine (or even: none) - think of command-line browsers as an example.

I've updated the patch, so please re-review.

index.bs
```diff
@@ -4125,22 +4129,20 @@ follows:</p>
...
 
-<p class="note">This section describes in some detail how to visually render <a>WebVTT caption or
-subtitle cues</a> in a user agent. The processing model is quite tightly linked to media elements in
-HTML, where CSS is available. <a>User agents that do not support CSS</a> are expected to render
-plain text only, without styling and positioning features. <a>User agents that do not support a full
-HTML CSS engine</a> are expected to render an equivalent visual representation to what a user agent
-with a full CSS engine would render.</p>
+<p>This section describes how to visually render WebVTT cues in a user agent. The processing model
+is tightly linked to media elements in HTML and follows the handling of cues when <a>playing the
+media resource</a>. When supporting WebVTT in media players that falls into the <a>User agents that
```
nigelmegitt: Grammar nit: s/in media players/in a media player

index.bs
```diff
@@ -4125,22 +4129,20 @@ follows:</p>
...
+do not support CSS</a> or <a>User agents that do not support a full HTML CSS engine</a> conformace
```
nigelmegitt: typo: s/conformace/conformance

index.bs
```diff
@@ -4125,22 +4129,20 @@ follows:</p>
...
+classes, equivalent visual rendering will need to be implemented.</p>
```
nigelmegitt: Another nit: s/will need/needs

Issue #375, opened 2017-09-27
Wide Review Comment 2017: position and size reference is video or video viewport?


nigelmegitt:
Copy/paste from https://lists.w3.org/Archives/Public/public-tt/2017Sep/0080.html - raising as an issue for tracking/disposition purposes.

In some places sizes and positions are defined relative to the video viewport; in others the video itself. This is likely to cause some confusion or mis-alignment when the two are not the same (e.g. a 16:9 aspect ratio video is displayed in a 14:9 viewport) and creates an authoring problem. For example the cue box size is relative to the video but the cue box line is relative to the video viewport (both defined within section 3.1).

nigelmegitt:
Additional point on this - there are some places where the `vw` and `vh` CSS units are used for positioning and sizing, yet they do not seem to accommodate scenarios where video is not shown full-window. Note that `vw` and `vh` are always relative to the initial containing block. 

I'm happy to be shown otherwise but I don't believe that CSS currently has a concept for setting a local viewport within which the `vw` and `vh` units are resolved. This means that the rule:

> The viewport (and initial containing block) is video’s rendering area.

does not actually seem to be implementable - I see that there's an open Chromium bug on this at https://bugs.chromium.org/p/chromium/issues/detail?id=244618 raised by people who found this a lot earlier than me.

silviapfeiffer:
About first comment: Well spotted - "a percentage of the video" means nothing - it should indeed be "video viewport".

About second comment: The way CSS is defined, vh and vw refer to the viewport, which is the "containing block" as per https://www.w3.org/TR/CSS21/visudet.html#containing-block-details . This also states that for media elements the "initial containing block" is the video viewport, so all WebVTT dimensions are calculated with respect to the block provided by the video. That is independent of whether the video is fullscreen or not - it's always relative to the actual rendered dimensions of the video. The sentence you quoted is merely repeating what the CSS spec says.

The bug that you are referring to simply says that Chromium's implementation of vh and vw for all shadow DOM elements of the video element (and that includes WebVTT rendering) is broken.

dwsinger:
are vw and vh established by the containing context (e.g. HTML) if any, or are they reset by the video element? If the latter, then it won't be possible to be consistent in e.g. font sizing. The phrase "The viewport (and initial containing block) is video’s rendering area." essentially sets a context for  this instantiation of CSS.  Perhaps it needs to say "The CSS viewport (as defined in CSS 2.1 section XX)..."

[...]

PR #420, opened 2017-12-08
Cue setting clarifications

silviapfeiffer:


index.bs
```diff
@@ -1903,10 +1903,10 @@ for vertical cues, it's the horizontal offset. The offset is for the <a lt="WebV
 alignment">start</a>, <a lt="WebVTT cue line center alignment">center</a>, or <a lt="WebVTT cue line
 end alignment">end</a> of the cue box, depending on the <a>WebVTT cue line alignment</a> value - <a
 lt="WebVTT cue line start alignment">start</a> by default. The offset can be given either as a
-percentage of the video dimension or as a line number. Line numbers are based on the size of the
-first line of the cue. Positive line numbers count from the start of the video viewport (the first
-line is numbered 0), negative line numbers from the end of the viewport (the last line is numbered
-&#x2212;1).</p>
+percentage of the relevant writing-mode dependent video dimension or as a line number. Line numbers
```
nigelmegitt: The video dimension not the video viewport dimension?

nigelmegitt:
Is there an issue reference for this pull request please?

index.bs
silviapfeiffer: The "video viewport" is the rendering area of the video, which is the only area that we're able to render anything onto, so for all intents and purposes it's really just a description of the video dimensions.

silviapfeiffer:
Hmm, I don't know where this was discussed now and forgot to link it in the commit. I think it might have been #375 but only addresses a part of it.

index.bs
silviapfeiffer: @nigelmegitt I can use "video viewport dimension" if you think it makes more sense.

nigelmegitt: @silviapfeiffer that would seem more consistent, modulo the comments on the related issue, that viewport units are not video viewport units.

silviapfeiffer: @nigelmegitt for now, I've made sure to use "video viewport" everywhere to clarify for this issue. I'll address the definition of the term "video viewport" in #375 .

nigelmegitt: OK, thanks @silviapfeiffer

PR #357, opened 2017-08-22
Ask for web-platform-tests in CONTRIBUTING.md

foolip:
The wording is the same as in:
https://github.com/w3c/IndexedDB/blob/master/CONTRIBUTING.md#tests
https://github.com/w3c/pointerlock/blob/gh-pages/CONTRIBUTING.md#tests
https://github.com/w3c/web-animations/blob/master/CONTRIBUTING.md#tests

Which was adapted from the following documents:
https://github.com/whatwg/meta/blob/master/CONTRIBUTING.md#tests
https://github.com/whatwg/meta/blob/master/TEAM.md
https://github.com/w3c/ServiceWorker/blob/master/CONTRIBUTING.md
https://github.com/w3c/web-performance/blob/gh-pages/CONTRIBUTING.md#test-driven

PR #421, opened 2017-12-21
Computed position alignment

### LINKED ISSUE #413: position, text_alignment issue
According  to spec "https://w3c.github.io/webvtt/"

problem 1:
=======
if cue contains "position:0% align:start"
https://w3c.github.io/webvtt/#webvtt-cue-position =>computed position is 0, 
https://w3c.github.io/webvtt/#webvtt-cue-position-alignment =>computed position alignment is center

https://w3c.github.io/webvtt/#processing-model
If the computed position alignment is center, and the computed position is less than or equal to 50
maximum_size = computed position * 2; which is 0

cue_size is minimum of (cue_size or maxium_size). From this cue_size is 0.
In spec it is not specified what should be done if cue_size is zero.

Problem 2:
=======
if cue contains "position:10% align:start"
https://w3c.github.io/webvtt/#webvtt-cue-position =>computed position is 10, 
https://w3c.github.io/webvtt/#webvtt-cue-position-alignment =>computed position alignment is center

https://w3c.github.io/webvtt/#processing-model
If the computed position alignment is center, and the computed position is less than or equal to 50
maximum_size = computed position * 2; which is 20

cue_size is minimum of (cue_size or maxium_size). From this cue_size is 20.
Determine the value of x-position for horizontal writing direction
If the computed position alignment is center alignment
Let x-position be the computed position minus half of size. => 10-20/2 =0;

Here position whichever specified in cue is not considered.
Same problem is there for
position> 50 && align:end;
position< 50 && align:start; cases

silviapfeiffer:
Closes #413

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,5 @@
+# Code of Conduct
+
+All code and communication in this repository are covered by the [W3C
+Code of Ethics and Professional
+Conduct](https://www.w3.org/Consortium/cepc/).
--- a/index.bs
+++ b/index.bs
@@ -1795,8 +1795,15 @@ replaced.</p>
 
 <h3 id=cue-settings>WebVTT cue settings</h3>
 
-<p>A <a>WebVTT cue settings list</a> consists of zero or more of the following settings. Each
-setting must not be included more than once per <a>WebVTT cue settings list</a>.</p>
+<p>A <a>WebVTT cue setting</a> is part of a <a>WebVTT cue settings list</a> and provides
+configuration options regarding the position and alignment of the cue box and the cue text
+within.</p>
+
+<p class="note">For example, a set of WebVTT cue settings may allow a cue box to be aligned to the
+left or positioned at the top right with the cue text within center aligned.</p>
+
+<p>The current available <a>WebVTT cue settings</a> that may appear in a <a>WebVTT cue settings
+list</a> are:</p>
 
 <ul class="brief">
  <li>A <a>WebVTT vertical text cue setting</a>.</li>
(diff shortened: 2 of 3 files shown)


zcorpan:
May I suggest deleting the master branch? ☺ seems like it's causing trouble still.

silviapfeiffer:
Will do - every now and then I accidentally merge master, but it doesn't really show in diff, so I didn't bother changing it. 😀

rakuco:
Hi there,

@foolip and I have been looking at [normative spec change without tests](https://docs.google.com/document/d/1kFYTqUMEbo9p87i6wAzORoicMWgHsuYwL2CgtvdhDaE/edit?usp=sharing) every week this quarter, and I noticed this seems to be the case here.

https://github.com/w3c/web-platform-tests/commit/5b4716ec395661786fa01692825a8f1b835ae328 appears to have added some tests for this behavior, but before the spec was updated.

I was wondering if there are any plans to update the tests to match this PR.

silviapfeiffer:
There's a couple of new tests missing for new commits and it's for two reasons (and entirely my fault): I have to learn to write the tests and I didn't want to hold up the standard process with these patches. I guess a new issue should be opened. What process are you following? Land patches only after tests have been prepared?

dwsinger:
Just like the spec., the test suite is a community effort. If you perceive a test as missing, inadequate, or wrong, please try to contribute the necessary improvements. It's not always the editor's job (nor is improved spec. text - pull requests for spec.-ready text are also appreciated).

foolip:
> What process are you following? Land patches only after tests have been prepared?

https://github.com/w3c/webvtt/blob/gh-pages/CONTRIBUTING.md#tests (added in https://github.com/w3c/webvtt/pull/357) is the same wording as we're using for all WHATWG specs and lots of W3C specs too. The way that seems to work best by some margin is actually to block the spec changes on the test changes.

Second to that, filing a issue on web-platform-tests that tests are missing, but it's quite likely those issues will sit idle unless they're part of someone's triage queue.

foolip:
I couldn't find everyone on GitHub, so I asked people on https://chromium-review.googlesource.com/c/chromium/src/+/756142 to take a look here. @fsoder I know though :)

fsoder:
The tests in the commit referenced above are not testing the spec change here (the code change isn't actually landed yet.) We have some existing tests that should cover this (what caused #413), that I hope will be exported when the code change lands.

foolip:
@fsoder, that's great! Upstreaming the tests with the implementation is really ideal, no extra work for the editors. Is the review for that already up, or a tracking bug?

Also found @k2nagaraju via https://github.com/w3c/webvtt/issues/413. FYI ^

dwsinger:
thanks everyone.  the recent clean-up efforts are basically done, so this is a good time to get the test suite and implementations all up to date and squeaky clean!

fsoder:
Yes the review is up, and essentially only missing said test updates (IIRC).

foolip:
I think https://chromium-review.googlesource.com/c/chromium/src/+/845462 is the review, certainly the code changes look like they match the spec changes in this PR.

Issue #379, opened 2017-09-27
Wide Review Comment 2017: unspecified edge margins make accurate positioning impossible


nigelmegitt:
Copy/paste from https://lists.w3.org/Archives/Public/public-tt/2017Sep/0080.html - raising as an issue for tracking/disposition purposes.

The user-agent defined edge margins used in section 6 Rendering effectively mean that the author of the file can not predict positions or line breaks accurately since the size and position of text cannot be known at authoring time. This prevents the author from being able to position captions in a location that avoids overlapping parts of the video that are important for the viewer to see in order to understand the content. This makes it effectively impossible to meet the BBC's [editorial subtitle guidelines](http://bbc.github.io/subtitle-guidelines/#Positioning) using WebVTT. Note that this comment must logically only apply when a single track is visible since WebVTT defines rules for multiple tracks being displayed simultaneously, in which case positioning is presumably outside the realm of the subtitle author, in the general case.

As an alternative to edge margins, why not allow the something like the padding property to be set on whatever viewport is used as the positioning reference area? (cf comment above about positioning relative to video or to video viewport) By the way, from a CSS box model perspective, calling them "margins" is confusing – aren't they "padding" values since they are inset from the video viewport area?

silviapfeiffer:
The edge margin was defined to address overscan. A captioning author has generally no idea if there is overscan in the video - the only entity involved in the steps necessary to render captions on top of video that knows whether there is overscan is the browser. Therefore, this gives the browser a chance to deal with overscan and allow the caption to be visible. E.g. when a caption author specifies that a cue is to be rendered at the left edge of the video, then with the left edge of the video being off screen, the cue may actually be partially invisible.

This is to deal with faulty authored caption files. I don't expect that would be the case with files coming from the BBC, so any explicit positioning provided in a BBC file would be rendered accurately.

Incidentally, the edge margin is exactly a kind of padding on the viewport. The way it is specified, it's a margin on the initial containing block for the captions rather than a padding on the viewport.
I'd be happy to add an explanatory sentence about how this influences the size of the initial containing block for the captions to address overscan. Would that be acceptable?

silviapfeiffer:
Note that CEA-608 defines a safe caption area as 80% of the TV picture height and 80% of TV picture width, centered. This is merely following in that footstep.

nigelmegitt:
> I'd be happy to add an explanatory sentence about how this influences the size of the initial containing block for the captions to address overscan. Would that be acceptable?

[...]

PR #422, opened 2017-12-21
Don't make edge margins depend on aesthetics.

silviapfeiffer:
This makes a start on #379

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4291,9 +4291,9 @@ must run the following algorithm.</p>
      <li><p>Let |edge margin| be a user-agent-defined horizontal length, expressed as a percentage
      of the width of the |video|'s rendering area, which will be used to define a margin at the left
      and right edges of the video into which this cue will not be placed. In situations with
-     overscan, this margin should be sufficient to place the cue within the title-safe area. In the
-     absence of overscan, this value should be picked for aesthetics (to avoid text being aligned
-     precisely on the left or right edge of the video, which can be ugly).</p></li>
+     overscan, this margin should be sufficient to place the cue within the title-safe area. Even in
+     the absence of overscan, it enhances readability to avoid text being aligned precisely on the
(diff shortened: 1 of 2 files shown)


silviapfeiffer:
wouldn't mind a review by @eric-carlson as I'm removing all mention of "edge margin" and "margin"

index.bs
```diff
@@ -1057,9 +1056,9 @@ fragment:</p>
 
   <ol algorithm="computed position">
 
-   <li><p>If the <a lt="WebVTT cue position">position</a> is numeric, then return the value of the
-   <a lt="WebVTT cue position">position</a> and abort these steps. (Otherwise, the <a lt="WebVTT cue
-   position">position</a> is the special value <a lt="WebVTT cue automatic
+   <li><p>If the <a lt="WebVTT cue position">position</a> is numeric between 0 and 100, then return
```
eric-carlson: I think this change is, strictly speaking, unrelated to this bug. If you include it, I suggest:

"... is **a** numeric **value** between 0 and 100 ..."

silviapfeiffer: Yes, that's true - it's a fly-by error I noticed as a consequence of checking the margin calculations. I could open another PR, but it's slightly related because it might cause wrong position calculations. If you're ok with it, I'll just fix the language as proposed.

eric-carlson: Sure, that is fine with me.

zcorpan:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=1012242 about this change.

silviapfeiffer:
OK, but this landed a long time ago...

zcorpan:
And it's still not implemented. This is why it's a good idea to have a process in place to update tests and file bugs when specs change, like the WHATWG does. 🙂 Otherwise it's like "if a tree falls in the forest..."

PR #423, opened 2017-12-22
Region lines cannot be negative now that they are unsigned long.

### LINKED ISSUE #414: The VTTRegion.lines attribute is a long type but cannot accept negative values
According to spec....
[region-lines](https://w3c.github.io/webvtt/#dom-vttregion-lines)
> The lines attribute, on getting, must return the WebVTT region lines of the WebVTT region that the VTTRegion object represents. On setting, if the new value is negative, then an IndexSizeError exception must be thrown. Otherwise, the WebVTT region lines must be set to the new value.

But according to webidl spec long type is ...
[long-type](https://heycam.github.io/webidl/#idl-long)
> The long type is a signed integer type that has values in the range [−2147483648, 2147483647].

So if we are not accepting negative values for lines, then how about changing its type to "unsigned long"?

silviapfeiffer:
Closes #414


k2nagaraju:
please merge this change.

silviapfeiffer:
@dwsinger I need a review

Issue #371, opened 2017-09-27
Wide Review Comment 2017: default font size and line heights


nigelmegitt:
Copy/paste from https://lists.w3.org/Archives/Public/public-tt/2017Sep/0080.html - raising as an issue for tracking/disposition purposes.

The default font size is 5vh, however the default line spacing for addressing line positions is only 5.33vh, representing a line spacing of 106.6% - that would represent an unusually small line height and may result in strange effects. I cannot see anywhere that overrides the line position height based on the computed value of the line-height property.

silviapfeiffer:
vh stands for viewport height. The line height of 5.33vh stems from CEA-608/708 which provide for 15 lines of content on 80% height, so that gives each line 5.33% height.

NOTE: 5.33vh is only used for regions, not for cues, where line-height is simply the CSS default "normal", which is typically about 1.2% larger.

So we just need to fix this for regions.
I did some forensics on how this happened.

1/ Originally, for regions, line height was fixed to 5.33vh and font-size to 5.33/1.3vh giving sufficient space. https://github.com/w3c/webvtt/pull/242 removed that because of https://www.w3.org/Bugs/Public/show_bug.cgi?id=25636 .
2/ We also still have #269 .

Now, how to move forward?
I'd rather not change the 5vh default font-size on cues because it's already implemented everywhere and seems to work quite well.
I think we could just replace the 5.33vh with "normal" or with 6vh (which is 5*1.2) to provide more space. That means we ignore the inspiration from CEA608/708, which is fine - if somebody needs it fixed to what CEA608/708 do, they can manually set the font-size and line-height through CSS in STYLE, which is what they have to do already.

Would 6vh work for you?

nigelmegitt:
OK, changing the default line height seems reasonable - I like the idea of clearly specified values so 6vh seems good, but I'm also aware that "normal" may give better results in terms of readability, so I don't have a strong position there.

Is there a place in the rendering algorithm where the line spacing for addressing line positions is obtained from the applicable style if set? I could not see any steps that adjusted _lineHeight_ from the initial 5.33vh value. Indeed it only appears in §6.1 subsection 12.1 bullet 2. So there appears to be no opportunity to change _lines_ based on an actual value.

silviapfeiffer:
We have a fixed line height for regions so we're able to calculate the region's height. _lineHeight_ is adjusted when applying the terms of the CSS.

I'll set the line height for regions to 6vh.

[...]

PR #430, opened 2018-03-30
Fontsize changes

silviapfeiffer:
Addresses https://github.com/w3c/webvtt/issues/371#issuecomment-350364574


index.bs
```diff
@@ -4219,6 +4219,8 @@ manner suiting the user.</p>
    <li>
     <p>Apply the terms of the CSS specifications to |regionNode| within the following constraints,
     thus obtaining a CSS box |box| positioned relative to an initial containing block:</p>
+    <p class="note">The CSS specification application may make changes to a number of variables,
+    including |lineHeight|.</p>
```
nigelmegitt: @silviapfeiffer am I right in thinking that those changes do _not_ then modify the height of a region by changing the value of _lines_ ? If so, I really think this needs to be stated, otherwise people might assume that the region has enough size to fit some number of lines but then after applying CSS to change lineHeight, find that it does not.

silviapfeiffer: We've been there, see https://github.com/w3c/webvtt/issues/371#issuecomment-346788337 , which is why we added this sentence. The CSS specification can change all those things including the height of the region.

nigelmegitt: Yes we have! Without wanting to go round in circles, as noted at https://github.com/w3c/webvtt/issues/371#issuecomment-351439545 adjusting the CSS line height does not affect the value of _lines_ because there seems to be no step that goes back to modify it. If there actually is a step, then it is non-obvious and it would be helpful to point to it in this note, but if there is indeed no such step (and we don't add one), then that also would be worth stating in this note.

silviapfeiffer: `lines` does not get changed, but the height of the region might because the `lineHeight` does. What would you like me to add? I'm very confused!

nigelmegitt: There's no mechanism for the height of the region to change based on `lineHeight` changing though, unless I've missed it? §7.1 Processing Model, bullet 12 step 1, bullet 2, sets `lineHeight` and `regionHeight` and derives the one time value `lines`. Then two bullets later the top is calculated before any CSS settings have been applied. Effectively at this point there is a fixed height. 

Then in bullet 12 step 2, there's the note introduced here that says that _lineHeight_ might be modified by CSS specification application. How is the region height then recalculated? 

Side-note: I couldn't find any spec text that maps the CSS `line-height` value to _lineHeight_. 

I'm reading the text "Let x be y" in an ES6-ish way, i.e. that the value of x is bound immediately and not re-evaluated whenever it is used.

So, IF my reasoning here is correct, then I'm asking for a sentence to be added to say that even if _lineHeight_ is changed due to CSS, the height of the region is not recalculated. OR, if my reasoning is incorrect (and please do explain), then add something like "which modifies the height of the region via XXX" where XXX is a pointer to a spec section or algorithm.

PR #431, opened 2018-03-30
Fix CR process link

silviapfeiffer:
Now pointing to Feb 2018 document


silviapfeiffer:
sorry, this got all confused - I'll re-do

PR #437, opened 2018-04-03
Clarify relationship between CR and CG report

### LINKED ISSUE #436: Clarify SOTD and abstract of proposed CR
Clarify that the CR is based on the CG report, and that it is a product of the TTWG.

palemieux:
Closes #436

### DIFF
--- a/archives/2018-04-15/Overview.html
+++ b/archives/2018-04-15/Overview.html
@@ -1509,6 +1509,10 @@ W3C <a href="http://www.w3.org/Consortium/Legal/ipr-notice#Legal_Disclaimer">lia
    <p>This specification defines WebVTT, the Web Video Text Tracks format. Its main use is for marking up external text track resources in connection with the HTML &lt;track> element.
 
 WebVTT files provide captions or subtitles for video content, and also text video descriptions <a data-link-type="biblio" href="#biblio-maur">[MAUR]</a>, chapters for content navigation, and more generally any form of metadata that is time-aligned with audio or video content.</p>
+
+<p>This specification is based on the <a href="https://w3c.github.io/webvtt/">Draft Community Group Report</a> of the
+<a href="https://www.w3.org/community/texttracks/">Web Media Text Tracks Community Group</a>.</p>
+
   </div>
   <h2 class="no-num no-toc no-ref heading settled" id="status"><span class="content">Status of this document</span></h2>
   <div data-fill-with="status">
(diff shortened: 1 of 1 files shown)


silviapfeiffer:
Can we get @tmichel07 input on this? What is the usual way this is done in the W3C?

nigelmegitt:
> not as clear about the relationship as I would like, but whatever

@dwsinger What aspects of the relationship do readers of the spec need to know about?

archives/2018-04-15/Overview.html
```diff
@@ -1571,13 +1567,6 @@ <h2 class="no-num no-toc no-ref heading settled" id="status"><span class="conten
 <p> This document is governed by the <a href="https://www.w3.org/2017/Process-20180201/" id="w3c_process_revision">1 February 2018 W3C Process Document</a>.</p>
 
   </div>
-  <div class="snapshot" hidden="">
-   <p>
-    This is a Recommendation Track snapshot. For the latest updates, possibly including important
- bug fixes, please see the <a href="https://w3c.github.io/webvtt/">Draft Community Group Report</a>. 
-    <button>Dismiss</button>
```
nigelmegitt: Not sure why the snapshot guidance needs to be removed; I agree with @silviapfeiffer we need W3 guidance on what is "normal" when managing specs in two places at once like this.

nigelmegitt: That's true. Plus I think the best forum for communicating about the spec is via GitHub repo issues, and anyone can do that whether they're in the CG, WG, both or neither.

silviapfeiffer: I propose we change
&lt;a href="https://w3c.github.io/webvtt/">Draft Community Group Report&lt;/a>
to
&lt;a href="https://w3c.github.io/webvtt/">Editor's Draft&lt;/a>

dwsinger: OK by me

dwsinger:
I think Silvia's change is fine. Let's do it. Getting the CR done is more important than finessing this more.

archives/2018-04-15/Overview.html
nigelmegitt: There should already be an Editor's Draft link at the top of the document. If the solution is to change the link to the ED then we should just remove the whole paragraph as per the pull request. Folk don't need to be told that the ED might be more up to date, that's obvious.

silviapfeiffer:
OK, let's merge it then.

PR #438, opened 2018-04-04
Update CLA and License agreement

palemieux:
The idea is for the CLA to require upfront a commitment from contributors that is compatible with Rec Track documents.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,26 +1,16 @@
-# Web Media Text Tracks Community Group
+# Contributor License Agreement
 
-This repository is being used for work in the Web Media Text Tracks Community Group, governed by the [W3C Community License
-Agreement (CLA)](http://www.w3.org/community/about/agreements/cla/). To contribute, you must join
-the CG.
+Contributions to this repository are intended to become part of Recommendation-track documents governed by the W3C Patent Policy and Document License. To contribute, you must either participate in the [Timed Text Working Group](https://www.w3.org/AudioVideo/TT/) or make a non-member patent licensing commitment.
 
-If you are not the sole contributor to a contribution (pull request), please identify all
-contributors in the pull request's body or in subsequent comments.
+If you are not the sole contributor to a contribution (pull request), please identify all contributors in the pull request's body or in subsequent comments.
 
 To add a contributor (other than yourself, that's automatic), mark them one per line as follows:
 
-```
 +@github_username
-```
-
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,26 +1,16 @@
-# Web Media Text Tracks Community Group
+# Contributor License Agreement
 
-This repository is being used for work in the Web Media Text Tracks Community Group, governed by the [W3C Community License
-Agreement (CLA)](http://www.w3.org/community/about/agreements/cla/). To contribute, you must join
```
silviapfeiffer: I disagree with removing the CLA and CG reference. It encourages people that cannot join the W3C to participate.

CONTRIBUTING.md
```diff
@@ -1,26 +1,16 @@
...
-the CG.
+Contributions to this repository are intended to become part of Recommendation-track documents governed by the W3C Patent Policy and Document License. To contribute, you must either participate in the [Timed Text Working Group](https://www.w3.org/AudioVideo/TT/) or make a non-member patent licensing commitment.
```
silviapfeiffer: You don't have to participate in the TTWG to make a contribution and the CLA above already makes a non-member patent licensing commitment.

There is far less legal conflict between the way that the CG works and the way the WGs work than you think. I disagree with making this a TTWG-only repository.

CONTRIBUTING.md
palemieux: As a W3C specification on the REC track, I do not see why WebVTT should be subject to a CLA different than other W3C specifications on the REC track. The CG can create a separate repo if it wants to work on something other then the specification on the REC track.

CONTRIBUTING.md
palemieux: > You don't have to participate in the TTWG to make a contribution and the CLA above already makes a non-member patent licensing commitment.

The same is true for IMSC: anyone can make a contribution, bound to the W3C CLA.

silviapfeiffer:
The non-member patent licensing requirements that you are requesting people to make is already sorted in the CG. Why would we make it harder for others to contribute when it has worked well in the past? Other Working Groups are embracing this style of working also. In fact, you might want to consider introducing this for other TTWG work also, seeing as some new work is happening in CGs.

silviapfeiffer:
The key here is to avoid creating forks, since that had caused trouble in other WGs. We should not go backwards on a model that had worked well and make it worse.

PR #444, opened 2018-06-12
Update web-platform-tests URLs

foolip:


foolip:
Yep, this is mostly so that if people copy bits of specs around when starting new specs, it'll be the new URL. So I updated all specs :)

silviapfeiffer:
Thanks for keeping this updated!

PR #451, opened 2019-04-05
add self as editor, move Silvia to former editor

gkatsev:


gkatsev:
@silviapfeiffer bikeshed doesn't accept something like "NICTA, CSIRO", specifically, looks like the comma is what it complains about. Some options:
- keep the two lines
- go with "NICTA CSIRO"
- keep only one line with an email that works

silviapfeiffer:
Go with "NICTA CSIRO" - or you could do  "NICTA/CSIRO"

PR #449, opened 2019-03-06
mark at-risk features in a new snapshot

gkatsev:
Made a new snapshot for 2019-03-06 and manually applied 2c06667 and c9100e0.
Then I marked collision avoidance with snap-to-lines false, `::cue-region` and `::cue-region()` pseudo-elements, and `:past` and `:future` pseudo-classes as at-risk.
STYLE block is no longer marked at risk because I have a [prototype](https://github.com/videojs/vtt.js/pull/35) that would be completed during the mandatory 28-day waiting period after this snapshot is published as the new CR.

PR #453, opened 2019-05-02
New snapshot with removal of at-risk items

gkatsev:
This is adds a new snapshot that can be used for a PR when moving forward. It removes the at-risk items.

Live overview: https://htmlpreview.github.io/?https://github.com/gkatsev/webvtt/blob/at-risk-snapshot/archives/2019-05-02/Overview.html
Live changes: https://htmlpreview.github.io/?https://github.com/gkatsev/webvtt/blob/at-risk-snapshot/archives/2019-05-02/changes.html
Live diff: https://htmlpreview.github.io/?https://github.com/gkatsev/webvtt/blob/at-risk-snapshot/archives/2019-05-02/diff.html


gkatsev:
This PR creates a new snapshot based on #449 which removes the at-risk items. I'm going to close this since there's now #460, #461, #470, #471, and #493 that we'd want to get in. Plus, with the new process, we may not need/want to remove these outright. Either way, I think a new one of these would be warranted.

PR #454, opened 2019-05-22
chore(ECHIDNA): use & instead of ;

marcoscaceres:
ReSpec no longer supports ; ... this fixes auto-publishing through ECHIDNA.


plehegar:
Marked as non substantive for IPR from ash-nazg.

PR #460, opened 2019-06-19
New snapshot with new at-risk items

gkatsev:
As discussed at t[he last TTWG call](https://www.w3.org/2019/06/13-tt-minutes.html#x02), marks text-combine-upright, text-wrap: balance, line alignment, and position alignment.

Fixes #455, fixes #456, includes #461 and #470.

Overview: https://htmlpreview.github.io/?https://github.com/gkatsev/webvtt/blob/at-risk-june/archives/2019-06-19/Overview.html
Changes: https://htmlpreview.github.io/?https://github.com/gkatsev/webvtt/blob/at-risk-june/archives/2019-06-19/changes.html
Diff: https://htmlpreview.github.io/?https://github.com/gkatsev/webvtt/blob/at-risk-june/archives/2019-06-19/diff.html

***
<a href="https://pr-preview.s3.amazonaws.com/gkatsev/webvtt/pull/460.html" title="Last updated on Sep 16, 2019, 7:09 AM UTC (70c0a9b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webvtt/460/5ddbe7b...gkatsev:70c0a9b.html" title="Last updated on Sep 16, 2019, 7:09 AM UTC (70c0a9b)">Diff</a>

PR #461, opened 2019-06-24
Make VTTRegion lines be a long.

### LINKED ISSUE #457: Change line attribute to unsigned int
as discussed on today's call

gkatsev:
Because WebIDL doesn't include an unsigned int, the VTTRegion interface
specifically calls out that negative numbers should cause an
IndexSizeError, and browser implementations are in the unsigned int,
switch over to using a regular long value.
Using a long gives us the correct range and with the definition that
specifies throwing means that negative values are not allowed.

This was discussed in the [20 June, 2019 TTWG call](https://www.w3.org/2019/06/20-tt-minutes.html#x02)

Fixes #457.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5926,7 +5926,7 @@ enum ScrollSetting { "" /* none */, "up" };
 interface VTTRegion {
   attribute DOMString id;
   attribute double width;
-  attribute unsigned long lines;
+  attribute long lines;
   attribute double regionAnchorX;
   attribute double regionAnchorY;
   attribute double viewportAnchorX;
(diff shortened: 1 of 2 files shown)


gkatsev:
Once this lands, #460 can be  updated with it and https://github.com/web-platform-tests/wpt/pull/17400 can be updated to remove the now unnecessary lines-long.html test.

silviapfeiffer:
Was the HTML auto-created using bikeshed?

gkatsev:
It was, I just cherry-picked only the relevant changes. I can make a separate update-bikeshed PR and then rebase that against it.

PR #470, opened 2019-07-15
Update region lines parsing to round to +/- MAX_VALUE

### LINKED ISSUE #467: [IR] parsing: region lines
The [region lines](https://github.com/web-platform-tests/wpt/blob/master/webvtt/parsing/file-parsing/tests/regions-lines.html) text parsing test is failing.
Some of the assertions pass but some fail due to the usage of very large number values. Safari is clamping the value to the max integer value while Firefox returns a 0. The spec just says to interpret the value as an integer but provides no guidance on it. This potentially means there's no way to test these edge cases. Though, may not be necessary.
Firefox's strict parser and Safari's non-standard implementations prevent the test from passing.

gkatsev:
This would potentially require an update to the WPT.

The text is borrowed from https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#rules-for-parsing-floating-point-number-values, though, I don't think that it's a great fit here.

I haven't regenerated the index.html yet, will do so after I figure out the process around bikeshed updates.

Fixes #467.

***
<a href="https://pr-preview.s3.amazonaws.com/gkatsev/webvtt/pull/470.html" title="Last updated on May 30, 2026, 4:26 PM UTC (4f11bc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webvtt/470/f6b3eab...gkatsev:4f11bc5.html" title="Last updated on May 30, 2026, 4:26 PM UTC (4f11bc5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2859,7 +2859,19 @@ Objects</a>.</p>
 
        <li><p>Interpret |value| as an integer, and let |number| be that number.</p></li>
 
-       <li><p>Let |region|'s <a>WebVTT region lines</a> be |number|.</p></li>
+       <li><p><i>Conversion</i>: Let <var>S</var> be the set of finite IEEE 754 double-precision
+       floating-point values except −0, but with two special values added:
+       <var>2<sup>1024</sup></var> and <var>−2<sup>1024</sup></var>.</p></li>
+
+       <li><p>Let |rounded-number| be the number in |S| that is closest to |number|,
+       selecting the number with an even significand if there are two equally close
+       values. (The two special values <var>2<sup>1024</sup></var> and <var>−2<sup>1024</sup></var>
+       are considered to have even significands for this purpose.)</p>
+
+       <li><p>If |rounded-number| is <var>2<sup>1024</sup></var> or <var>−2<sup>1024</sup></var>,
+       jump to the step labeled <i>next setting</i></p></li>
+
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2852,7 +2852,19 @@ Objects</a>.</p>
 
        <li><p>Interpret |value| as an integer, and let |number| be that number.</p></li>
 
-       <li><p>Let |region|'s <a>WebVTT region lines</a> be |number|.</p></li>
+       <li><p><i>Conversion</i>: Let <var>S</var> be the set of finite IEEE 754 double-precision
```
gkatsev: It's likely we don't need this `conversion` label

silviapfeiffer: It may be useful if we need to do this conversion elsewhere as well. Might be worth checking.

silviapfeiffer:
I would recommend adding this to the REC branch directly.

Then, for the master branch, update Bikeshed in a separate PR before making another PR with updates from the REC.

silviapfeiffer (approved):

gkatsev:
Yeah, I can create another bikeshed PR and then rebase this. I'll add it to #460 as well.
Should I fixup the bikeshed update as outlined in this commit? https://github.com/w3c/webvtt/commit/754f13e3cf03d6036c3e4628c6920d17b412f778 or is it possible it may not be necessary?

PR #478, opened 2019-12-16
Add repo-type to w3c.json

foolip:
This was found missing in https://w3c.github.io/validate-repos/report.html.


foolip:
@plehegar for review.

plehegar:
(looks like the ipr check was blocked, which could happen if the GH API has a hiccup. Since this PR isn't substantive, I went around the check...)

PR #492, opened 2020-11-02
Editorial: Align with Web IDL specification

autokagami:
This is an automated pull request to align the spec with the latest Web IDL specification.

Currently the autofix might introduce some awkward code formatting, so please feel free to modify the formatting.

Please file an issue on https://github.com/saschanaz/webidl-updater/issues/new if you think this PR is invalid or should be enhanced.

The following is the validation messages from webidl2.js, which may help understanding this PR:

```
Validation error at line 8 in webvtt1,0, inside `interface VTTCue`:
 Constructor(double startTime,
 ^ Constructors should now be represented as a `constructor()` operation on the interface instead of `[Constructor]` extended attribute. Refer to the [WebIDL spec section on constructor operations](https://heycam.github.io/webidl/#idl-constructors) for more information.

Validation error at line 3 in webvtt1,1, inside `interface VTTRegion`:
 Constructor]
 ^ Constructors should now be represented as a `constructor()` operation on the interface instead of `[Constructor]` extended attribute. Refer to the [WebIDL spec section on constructor operations](https://heycam.github.io/webidl/#idl-constructors) for more information.
```

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 26, 2021, 6:10 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwebvtt%2Ff5c28140f619a6f7cd301345515785ced4076d41%2Findex.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 2.
WARNING: The following &lt;var>s were only used once in the document:
  'stylesheets', in algorithm 'WebVTT parser algorithm'
  'input', in algorithm 'WebVTT region settings parsing'
  'input', in algorithm 'WebVTT region objects'
  'regions', in algorithm 'collect WebVTT cue timings and settings'
  'input', in algorithm 'parse the WebVTT cue settings'
  'regions', in algorithm 'parse the WebVTT cue settings'
  'language', in algorithm 'WebVTT cue text parsing rules'
[...]

PR #493, opened 2020-11-04
Added unbounded TextTrackCue.endTime

rjksmith:
Added support for unbounded TextTrackCue - see https://github.com/whatwg/html/pull/5953
Whitespace removed by Atom

***
<a href="https://pr-preview.s3.amazonaws.com/away-team-software/webvtt/pull/493.html" title="Last updated on Apr 28, 2021, 4:01 PM UTC (0f46a24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webvtt/493/6befe1d...away-team-software:0f46a24.html" title="Last updated on Apr 28, 2021, 4:01 PM UTC (0f46a24)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5569,7 +5569,8 @@ interface VTTCue : TextTrackCue {
  <dd>
   <p>Returns a new {{VTTCue}} object, for use with the {{TextTrack/addCue()}} method.</p>
   <p>The |startTime| argument sets the <a>text track cue start time</a>.</p>
-  <p>The |endTime| argument sets the <a>text track cue end time</a>.</p>
+  <p>The |endTime| argument sets the <a>text track cue end time</a>. In the case of the value being
+  positive Infinity, the {{VTTCue}} object represents an <a>unbounded text track cue</a>.</p>
   <p>The |text| argument sets the <a>cue text</a>.</p>
  </dd>
 
--- a/index.bs
+++ b/index.bs
@@ -54,6 +54,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: rules for extracting the chapter title
             text: text track cue start time
             text: text track cue end time
+            text: unbounded text track cue
             text: expose a user interface to the user
             text: text track cue order
             text: honor user preferences for automatic text track selection
@@ -5544,7 +5545,8 @@ enum DirectionSetting { "" /* horizontal */, "rl", "lr" };
 enum LineAlignSetting { "start", "center", "end" };
 enum PositionAlignSetting { "line-left", "center", "line-right", "auto" };
 enum AlignSetting { "start", "center", "end", "left", "right" };
-[Exposed=Window]
+[Exposed=Window,
+ Constructor(double startTime, unrestricted double endTime, DOMString text)]
 interface VTTCue : TextTrackCue {
   constructor(double startTime, double endTime, DOMString text);
   attribute VTTRegion? region;
@@ -5693,7 +5695,7 @@ interface VTTCue : TextTrackCue {
  interpreted as a time in seconds.</p></li>
 
  <li><p>Let |cue|'s <a>text track cue end time</a> be the value of the |endTime| argument,
- interpreted as a time in seconds.</p></li>
+ interpreted as a time in seconds or positive Infinity for an <a>unbounded text track cue</a>.</p></li>
 
  <li><p>Let |cue|'s <a>cue text</a> be the value of the |text| argument, and let the <a>rules for
  extracting the chapter title</a> be the <a>WebVTT rules for extracting the chapter
--- a/index.bs
+++ b/index.bs
@@ -5548,7 +5548,7 @@ enum AlignSetting { "start", "center", "end", "left", "right" };
 [Exposed=Window,
  Constructor(double startTime, unrestricted double endTime, DOMString text)]
 interface VTTCue : TextTrackCue {
-  constructor(double startTime, double endTime, DOMString text);
+  constructor(double startTime, unrestricted double endTime, DOMString text);
   attribute VTTRegion? region;
   attribute DirectionSetting vertical;
   attribute boolean snapToLines;
@@ -5734,6 +5734,13 @@ interface VTTCue : TextTrackCue {
  <li><p>Return the {{VTTCue}} object representing |cue|.</p></li>
 
 </ol>
+
+<p>The <dfn attribute for=VTTCue>endTime</dfn> attribute, on getting, must return the <a>text track
+cue end time</a> of the <a>WebVTT cue</a> that the {{VTTCue}} object represents, in seconds or
+positive Infinity. On setting, if the new value is negative Infinity or a Not-a-Number (NaN) value,
+then throw a <a
+href="https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror">TypeError
+</a> exception. Otherwise, the <a>text track cue end time</a> must be set to the new value.</p>
 
 <p>The <dfn attribute for=VTTCue>region</dfn> attribute, on getting, must return the {{VTTRegion}}
 object representing the <a>WebVTT cue region</a> of the <a>WebVTT cue</a> that the {{VTTCue}} object
--- a/index.bs
+++ b/index.bs
@@ -5545,8 +5545,7 @@ enum DirectionSetting { "" /* horizontal */, "rl", "lr" };
 enum LineAlignSetting { "start", "center", "end" };
 enum PositionAlignSetting { "line-left", "center", "line-right", "auto" };
 enum AlignSetting { "start", "center", "end", "left", "right" };
-[Exposed=Window,
- Constructor(double startTime, unrestricted double endTime, DOMString text)]
+[Exposed=Window]
 interface VTTCue : TextTrackCue {
   constructor(double startTime, unrestricted double endTime, DOMString text);
   attribute VTTRegion? region;
@@ -5694,8 +5693,11 @@ interface VTTCue : TextTrackCue {
  <li><p>Let |cue|'s <a>text track cue start time</a> be the value of the |startTime| argument,
  interpreted as a time in seconds.</p></li>
 
- <li><p>Let |cue|'s <a>text track cue end time</a> be the value of the |endTime| argument,
- interpreted as a time in seconds or positive Infinity for an <a>unbounded text track
+ <li><p>If the value of the |endTime| argument is negative Infinity or a Not-a-Number (NaN) value,
+ then throw a <a
+ href="https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror">TypeError
+ </a> exception. Otherwise, let |cue|'s <a>text track cue end time</a> be the value of the |endTime|
+ argument, interpreted as a time in seconds or positive Infinity for an <a>unbounded text track
  cue</a>.</p></li>
 
  <li><p>Let |cue|'s <a>cue text</a> be the value of the |text| argument, and let the <a>rules for
@@ -5734,13 +5736,6 @@ interface VTTCue : TextTrackCue {
  <li><p>Return the {{VTTCue}} object representing |cue|.</p></li>
 
 </ol>
-
-<p>The <dfn attribute for=VTTCue>endTime</dfn> attribute, on getting, must return the <a>text track
-cue end time</a> of the <a>WebVTT cue</a> that the {{VTTCue}} object represents, in seconds or
-positive Infinity. On setting, if the new value is negative Infinity or a Not-a-Number (NaN) value,
-then throw a <a
-href="https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror">TypeError
-</a> exception. Otherwise, the <a>text track cue end time</a> must be set to the new value.</p>
 
 <p>The <dfn attribute for=VTTCue>region</dfn> attribute, on getting, must return the {{VTTRegion}}
 object representing the <a>WebVTT cue region</a> of the <a>WebVTT cue</a> that the {{VTTCue}} object
--- a/index.bs
+++ b/index.bs
@@ -5690,15 +5690,14 @@ interface VTTCue : TextTrackCue {
 
  <li><p>Create a new <a>WebVTT cue</a>. Let |cue| be that <a>WebVTT cue</a>.</p></li>
 
- <li><p>Let |cue|'s <a>text track cue start time</a> be the value of the |startTime| argument,
- interpreted as a time in seconds.</p></li>
+ <li><p>Let |cue|'s <a>text track cue start time</a> be the value of the |startTime|
+ argument.</p></li>
 
  <li><p>If the value of the |endTime| argument is negative Infinity or a Not-a-Number (NaN) value,
  then throw a <a
  href="https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror">TypeError
  </a> exception. Otherwise, let |cue|'s <a>text track cue end time</a> be the value of the |endTime|
- argument, interpreted as a time in seconds or positive Infinity for an <a>unbounded text track
- cue</a>.</p></li>
+ argument.</p></li>
 
  <li><p>Let |cue|'s <a>cue text</a> be the value of the |text| argument, and let the <a>rules for
  extracting the chapter title</a> be the <a>WebVTT rules for extracting the chapter
--- a/index.bs
+++ b/index.bs
@@ -5695,7 +5695,8 @@ interface VTTCue : TextTrackCue {
  interpreted as a time in seconds.</p></li>
 
  <li><p>Let |cue|'s <a>text track cue end time</a> be the value of the |endTime| argument,
- interpreted as a time in seconds or positive Infinity for an <a>unbounded text track cue</a>.</p></li>
+ interpreted as a time in seconds or positive Infinity for an <a>unbounded text track
+ cue</a>.</p></li>
 
  <li><p>Let |cue|'s <a>cue text</a> be the value of the |text| argument, and let the <a>rules for
  extracting the chapter title</a> be the <a>WebVTT rules for extracting the chapter


rjksmith:
@nigelmegitt @gkatsev I was [asked to raise this PR](https://github.com/whatwg/html/issues/5297#issuecomment-714406798) to maintain `VTTCue` alignment with the proposal for `TextTrackCue.endTime = +Infinity`. The change was previously discussed in issue https://github.com/whatwg/html/issues/5297 and has now progressed to PR https://github.com/whatwg/html/pull/5953.

I now see that there's an [IPR issue](https://labs.w3.org/repo-manager/pr/id/w3c/webvtt/493) as I'm not a member of TTWG, though I am an IE in [SDWIG](https://www.w3.org/groups/ig/sdw/participants) and have been working with [MTE TF](https://github.com/w3c/me-media-timed-events) for a while. There's no urgency at the moment, but I'd like to resolve this blocking issue and would welcome your guidance on how best to proceed. Thanks.

cc @chrisn

tidoust:
@rjksmith, `index.html` is automatically generated from `index.bs`, using [Bikeshed](https://github.com/tabatkins/bikeshed#bikeshed-a-spec-preprocessor). In other words, the PR should update `index.bs`, not `index.html`.

Regarding the IPR issue, an easy way to solve this would be to have you sign a non-participant contribution commitment.

gkatsev:
Thanks @tidoust!

I think updating index.bs` would also trigger the PR Preview feature.

rjksmith:
@tidoust @gkatsev Thanks for your guidance.

I've reverted my changes (7e4b64c) to `index.html` and updated `index.bs` to generate equivalent modifications with Bikeshed (17626fd).

I suspect that the Bikeshed tool has been updated since `index.html` was last officially generated and I've split these changes into a separate commit (f0323cb) for clarity.

Please let me know the process to sign a non-participant contribution agreement to cover IPR so we can resolve the blocking issue. Thanks.

himorin:
Error is from WebIDL change at https://github.com/heycam/webidl/pull/798, which is supposed to be fixed by https://github.com/w3c/webvtt/pull/492

dontcallmedom:
@himorin let me know if you need any assistance with the non-participant contribution agreement  process - it should be relatively well-identified starting from https://labs.w3.org/repo-manager/pr/id/w3c/webvtt/493

himorin:
> @himorin let me know if you need any assistance with the non-participant contribution agreement process

ah, yes. For now, this is kept stop for repo management side (reorg, webidl alignment...), but should be fixed shortly.

gkatsev (approved):
The change is good to go, once we fix up the WebIDL issue we'll circle back about the IPR stuff and get this change in.

gkatsev:
We've updated the `main` branch, this PR would need to be updated/rebased against the latest changed there.

rjksmith:
[WebVMT already handles unbounded cues](https://w3c.github.io/sdw/proposals/geotagging/webvmt/#webvmtcues) as they are a common use case for live streaming.

The cue _end time_ is simply omitted as in the [Tower of London example](https://w3c.github.io/sdw/proposals/geotagging/webvmt/#mappresentation). This has the advantage that an _end time_ can easily be added to make the cue bounded when it becomes known at some future time.

gkatsev:
@rjksmith yeah, I brought up how WebVMT does it.
For WebVTT there the API and the format. This PR currently only changes the API for VTTCue, which is necessary for the programmatic metadata cues as notes in the examples. It probably makes sense for WebVTT to be able to represent it in its text format too, but then it must do so in a backwards compatible manner, which is what (2) in the summary above refers to.
Though, I think a syntax change can come in separately from the API change.

For the tests, we'd want to make a PR against web-platform-tests to update the constructor to allow endTime to be Infinity https://github.com/web-platform-tests/wpt/blob/master/webvtt/api/VTTCue/constructor.html.

gkatsev:
Oh, just realized that maybe TextTrackCue has tests too and sure enough it does. https://github.com/web-platform-tests/wpt/blob/master/html/semantics/embedded-content/media-elements/interfaces/TextTrackCue/endTime.html I'll add a comment on the html repo PR too.

gkatsev:
Also, I just verified that adopting WebVMT's syntax for unbounded cues *can* happen in a backwards compatible way. Parsers that don't recognize a time header without an end time (i.e. all current webvtt parsers) ignore that entire cue block but still show all subsequent ones. Tested in Chrome, Firefox, Safari, and vtt.js.

rjksmith:
> Also, I just verified that adopting WebVMT's syntax for unbounded cues _can_ happen in a backwards compatible way. Parsers that don't recognize a time header without an end time (i.e. all current webvtt parsers) ignore that entire cue block but still show all subsequent ones. Tested in Chrome, Firefox, Safari, and vtt.js.

@gkatsev Thanks for confirming. That sounds encouraging. I'll take a look at the web-platform-tests too.

rjksmith:
> > @himorin let me know if you need any assistance with the non-participant contribution agreement process
> 
> ah, yes. For now, this is kept stop for repo management side (reorg, webidl alignment...), but should be fixed shortly.

Is there any news on this or can you post a link to the agreement? Thanks

himorin:
> > > @himorin let me know if you need any assistance with the non-participant contribution agreement process
> > 
> > 
> > ah, yes. For now, this is kept stop for repo management side (reorg, webidl alignment...), but should be fixed shortly.
> 
> Is there any news on this or can you post a link to the agreement? Thanks

I think you might have received an email from W3C non-participant contribution system at Mar 11, 2021, to an email address you registered to W3C account.

gkatsev:
> @gkatsev Thanks for confirming. That sounds encouraging. I'll take a look at the web-platform-tests too.

It would be interesting to know whether it's useful to have a way to represent unbound end-time in a way that old parsers handle things intelligently. But it may not be. I think this was discussed a bit at the M&EIG meeting that, unfortunately, had to miss.

nigelmegitt:
> Also, I just verified that adopting WebVMT's syntax for unbounded cues _can_ happen in a backwards compatible way. Parsers that don't recognize a time header without an end time (i.e. all current webvtt parsers) ignore that entire cue block but still show all subsequent ones. Tested in Chrome, Firefox, Safari, and vtt.js.

I'm not convinced that skipping the entire cue counts as "backwards compatible". I'd suggest a more backwards compatible approach here is one where a cue end time is specified, that works for all parsers, and additional syntax is available, that is ignored by parsers that don't recognise it, and used by those that do, where that syntax has the semantic "set the end time to infinity". 

Then authors can specify a fallback end time of their choosing.

I'd propose a new WebVTT Cue Setting to satisfy this, because unknown cue settings should be ignored by parsers that don't recognise them.

For example:

```
00:30.000 --> 99:59.999 end-time-override: infinity
This cue is supposed to last effectively forever.
```

Current parser generates a cue whose end time is `99:59.999`.
Parser that understands `end-time-override: infinity` generates a cue whose end time is `Infinity`.

gkatsev:
I think it counts as backwards compatible, but it's definitely not the best behavior. I was verifying whether the current syntax that WebVMT used could be used and not wreak havoc on existing parsers, like having it fail to parse the rest of the file or something.
Behavior-wise, a new option definitely fits better with graceful degradation, which is definitely better than all or nothing. (For a similar reason, WebVTT really needs to add the `rp` element for ruby parentheticals).

rjksmith:
> I think you might have received an email from W3C non-participant contribution system at Mar 11, 2021, to an email address you registered to W3C account.

@himorin Many thanks. I've found it.

rjksmith:
Using unbounded cues is not compulsory. If backward compatibility is an issue then a long bounded cue can be used instead.
```
00:30.000 --> 99:59.999
This cue is supposed to last effectively forever.
```
This is fully backward compatible and is the current solution.

WebVMT also allows unbounded cues to be superseded which is an integral part of [interpolation for live streaming](https://github.com/webvmt/community-group/issues/2#issuecomment-708529659) use cases as [discussed at TPAC 2020](https://www.w3.org/2020/10/TPAC/breakout-schedule.html#video-location).

nigelmegitt:
> Using unbounded cues is not compulsory.

Indeed, but this issue is to allow the semantic of an unbounded cue to be expressed, right?

dwsinger:
I think that Nigel's point is that it's hard to tell for all cases what the best backwards-compatible behavior should be:
1. If you can't display it indefinitely, don't display it at all.
2. If you can't display it indefinitely, display it for a really long but finite time.

However, I am not sure what the effect of (2) would be. Experiment is needed. It might cause players to conclude that the presentation really is 99 years long, and display e.g. a scroll-bar that long.

rjksmith:
Yes I agree. However, displaying an unbounded cue indefinitely may not be the correct behaviour.

An unbounded cue is defined as a cue with an _unspecified_ future end time. The current WHATWG spec represents the duration of an unbounded stream with Infinity: [media.duration = Infinity](https://html.spec.whatwg.org/multipage/media.html#offsets-into-the-media-resource). The stream demonstrably has a finite duration, so Infinity is being used to represent _an unspecified future end time_ and the same definition applies to an unbounded cue - [as previously discussed](https://github.com/whatwg/html/issues/5297#issuecomment-652516934).

This can be illustrated with a simple use case. For example, the score during a live sporting event is "0-0" at the start of the game. This may change to "1-0" or "0-1" at an unknown future time or may stay the same, and the duration of the match may be extended. There is no way of knowing the duration of the "0-0" cue until the score changes or the game ends, so this is an unbounded cue.

The point is that it is not possible to predict the end time of an unbounded cue (or stream) because it is unspecified by definition, which is reflected in the [WebVMT cue syntax](https://w3c.github.io/sdw/proposals/geotagging/webvmt/#webvmtcues):
```
00:30.000 -->
This cue is unbounded
```
This is a new HTML feature and if there is a requirement for backward compatibility, bounded cues can be used instead to ensure the desired result.

dwsinger:
Um, the example you give doesn't seem to work, unless there is some provision to update a cue that's already been issued, decoded and acted on. The score 0-0 will last forever, even after a goal is scored.

Using captions for state is, I think, problematic.

nigelmegitt:
@dwsinger I agree - there has been talk (in the MEIG) of creating a way to update the end time of a cue, but no proposal so far for how to do it. One of the problems is how to identify which cue's end time needs to be updated, given that the uniqueness of any cue identifier is only constrained within a single WebVTT file.

If I understand correctly there has been some consideration of handling unbounded end times in the context of ISOBMFF? I wonder how the group looking at that imagined this working - can you tell us anything more?

rjksmith:
Yes it requires an update mechanism which exists in the HTML DOM. I agree that WebVTT would require modification to support this too.

WebVMT is designed to handle this issue for moving object trajectories, regions and sensors. Perhaps there's a way to apply this to text cues too.

dwsinger:
In the file format, we're merely addressing how to handle cues that are marked as indefinite; any update of end-time would be something we'd reflect if it were in the base spec.; I don't think we want to invent provisions at the MP4 level, merely encapsulate.

nigelmegitt:
> we're merely addressing how to handle cues that are marked as indefinite

I don't quite follow: there is no current way to mark a cue as indefinite in a file, so what is it that needs to be addressed?

dwsinger:
I mean, we're considering how to address cues that have an unbounded end time, which is this pull.

nigelmegitt:
> I mean, we're considering how to address cues that have an unbounded end time, which is this pull.

I'd be interested in @gkatsev 's views here, but it feels like we'll keep spinning until we have a clearer notion of what the use cases and requirements are, and if there are different groups hoping to get something out of the notion of an unbounded end time, they should probably sit down together in some way and share what their goals are.

dwsinger:
My impression was that this was settled and likely to land soon, and since ISO moves really slowly, we should start on the encapsulation work. If that's not the case, let me know. (Though text on "how to handle a wombat invasion" when in fact, there is no provision for wombats at all, is harmless).

nigelmegitt:
I'm more concerned that there could be conceptual differences in the models in our heads for how this might work, so there could indeed be harm - some people might be trying to handle a wombat invasion, and others might be trying to work out how to house wombats in a safe way, and nobody is asking the wombats.

gkatsev:
So, I think there are two things here.

1. update VTTCue API to allow Infinity for the endTime. This aligns it with the outstanding PR against TextTrackCue (https://github.com/whatwg/html/pull/5953) and comes from this issue https://github.com/whatwg/html/issues/5297.
Based on the original issue, I think this is mostly for request programmatic access and is mostly agreed upon, afaik.

(I was actually trying to test whether you can update a VTTCue live and how it updates but running into weird bugs, I'll try again another day)

2. As part of this PR, in the TTWG discussion, the question came up whether we wanted/needed syntax in WebVTT to represent this, and I think this is where a lot of the issues are coming up with. I think it would definitely be valid to decide that WebVTT can't represent unbounded cues, though, given that WebvTT is supposed to be also used for metadata, I think it probably should but can and should be addressed separately from this PR.

rjksmith:
One solution is to update the end time of an unbounded cue by superseding it with a bounded cue which has matching start time and content.

For example:
```
NOTE Unbounded cue

00:30.000 -->
This cue lasts until...

NOTE Subsequent cues go here

00:40.000 --> 00:50.000
This is a later cue

NOTE When known, update the unbounded cue end time

00:30.000 --> 01:30.000
This cue lasts until...

NOTE Cues after update time go here

01:30.000 --> 01:35.000
...now.
```
This would not require an identifier and has the advantage of including a valid bounded cue in the stream at the earliest moment when all its attributes are known - for compatibility. @gkatsev has already confirmed that [unbounded cues will be ignored by current WebVTT parsers](https://github.com/w3c/webvtt/pull/493#issuecomment-795777639). WebVTT cue time ordering would need to be relaxed for this exceptional case and could be used as an efficient mechanism for identifying updates.

gkatsev:
At FOMS in previous years we discussed how WebVTT should be used for live and segmented vtt files, particularly around cues that span multiple segments. One of the things that came out there is that you could match up cues with the same text and start time across segments and update the end time to that of the last cue that you got, kind of like @rjksmith shows above. This actually also fits well with the unbounded cue with the missing end time thing.

eric-carlson:
We already have the situation where a cue's end time is "unbounded" at first and is updated later, because some in-band text track formats deliver cues with only a start time. In these formats a cue ends when the next cue or empty edit is delivered. 

We support this in WebKit by giving these cues an end time of positive-infinity when they are delivered, and updating the end time when it is known

nigelmegitt:
> We already have the situation where a cue's end time is "unbounded" at first and is updated later, because some in-band text track formats deliver cues with only a start time. In these formats a cue ends when the next cue or empty edit is delivered.

That works in a semantic model where only one "cue" can be active at once, but that WebVTT's model does not enforce that constraint, so I don't understand how it helps here?

eric-carlson:
> > We already have the situation where a cue's end time is "unbounded" at first and is updated later, because some in-band text track formats deliver cues with only a start time. In these formats a cue ends when the next cue or empty edit is delivered.
> 
> That works in a semantic model where only one "cue" can be active at once, but WebVTT's model does not enforce that constraint, so I don't understand how it helps here?

I was merely pointing out that cues are being delivered on the web today whose duration is unknown when they are delivered, in case this isn't widely known.

rjksmith:
> We already have the situation where a cue's end time is "unbounded" at first and is updated later, because some in-band text track formats deliver cues with only a start time. In these formats a cue ends when the next cue or empty edit is delivered.
>
> We support this in WebKit by giving these cues an end time of positive-infinity when they are delivered, and updating the end time when it is known

@eric-carlson Thanks for your constructive feedback. Highlighting another valid use case is helpful and this could also be supported by WebVTT using the [syntax proposed above](https://github.com/w3c/webvtt/pull/493#issuecomment-808408634). WebVMT has a [similar use case](https://github.com/webvmt/community-group/issues/2#issuecomment-708529659) for live streaming.

nigelmegitt:
> Highlighting another valid use case is helpful and this could also be supported by WebVTT using the [syntax proposed above](https://github.com/w3c/webvtt/pull/493#issuecomment-808408634).

@rjksmith There are a lot of unanswered questions implied by the syntax proposal. First on my list is: given that WebVTT does not prohibit multiple cues with the same begin time, begin time seems to be unsuitable for use as a key to match the cue. How would you resolve this?

rjksmith:
> So, I think there are two things here.
> 
>    1. update VTTCue API to allow Infinity for the endTime. This aligns it with the outstanding PR against TextTrackCue ([whatwg/html#5953](https://github.com/whatwg/html/pull/5953)) and comes from this issue [whatwg/html#5297](https://github.com/whatwg/html/issues/5297).
>        Based on the original issue, I think this is mostly for request programmatic access and is mostly agreed upon, afaik.
> 
>    1. As part of this PR, in the TTWG discussion, the question came up whether we wanted/needed syntax in WebVTT to represent this, and I think this is where a lot of the issues are coming up with. I think it would definitely be valid to decide that WebVTT can't represent unbounded cues, though, given that WebvTT is supposed to be also used for metadata, I think it probably should but can and should be addressed separately from this PR.

@gkatsev I agree.
   1. There is a proposal to include unbounded cues in HTML TextTrackCue (https://github.com/whatwg/html/pull/5953) and this Pull Request addresses the minimum update required to WebVTT as a direct result of that change.
   1. Four use cases have been highlighted above - [WebKit](https://github.com/w3c/webvtt/pull/493#issuecomment-808429391), [FOMS](https://github.com/w3c/webvtt/pull/493#issuecomment-808411871), [0-0](https://github.com/w3c/webvtt/pull/493#issuecomment-808339093) and [WebVMT](https://github.com/w3c/webvtt/pull/493#issuecomment-808359344) - which relate to unbounded cues so there is an identified requirement which WebVTT could address in future. Syntax has been proposed to [represent unbounded cues in WebVTT](https://github.com/w3c/webvtt/pull/493#issuecomment-808408634) which demonstrates feasibility for inclusion, though I'm sure this first draft could be improved. These issues should be explored constructively as inputs to any decision to include unbounded cue support in WebVTT and are beyond the scope of this Pull Request. I'll raise an issue for unbounded cue use cases to progress that discussion separately and decouple it from this PR.

rjksmith:
> > Highlighting another valid use case is helpful and this could also be supported by WebVTT using the [syntax proposed above](https://github.com/w3c/webvtt/pull/493#issuecomment-808408634).
> 
> @rjksmith There are a lot of unanswered questions implied by the syntax proposal. First on my list is: given that WebVTT does not prohibit multiple cues with the same begin time, begin time seems to be unsuitable for use as a key to match the cue. How would you resolve this?

@nigelmegitt I agree that using _only_ start time to match the cue is unsuitable which is why I proposed matching by start time and content [as stated previously](https://github.com/w3c/webvtt/pull/493#issuecomment-808408634).

[As promised above](https://github.com/w3c/webvtt/pull/493#issuecomment-811810387), I've now raised [issue #496 for discussion of use cases and syntax changes](https://github.com/w3c/webvtt/issues/496) separately and suggest we continue this conversation there. Thank you.

rjksmith:
> > I think you might have received an email from W3C non-participant contribution system at Mar 11, 2021, to an email address you registered to W3C account.
> 
> @himorin Many thanks. I've found it.

Thanks for your help. I can confirm that I've now signed this.

rjksmith:
> Oh, just realized that maybe TextTrackCue has tests too and sure enough it does. https://github.com/web-platform-tests/wpt/blob/master/html/semantics/embedded-content/media-elements/interfaces/TextTrackCue/endTime.html I'll add a comment on the html repo PR too.

@gkatsev Having written suitable tests in web-platform-tests/wpt#28394, I noticed that https://github.com/web-platform-tests/wpt/blob/master/html/semantics/embedded-content/media-elements/interfaces/TextTrackCue relies on the VTTCue constructor so whatwg/html#5953 and #493 (this) are interdependent. I don't foresee a problem, but these changes need to be properly co-ordinated because of the Web Platform Tests. Hope this helps.

rjksmith:
@himorin I've signed the non-participant agreement but still see a [failing check](https://labs.w3.org/repo-manager/pr/id/w3c/webvtt/493) on this issue.

> Acceptable:  no
> Contributor: rjksmith needs to submit their non-participant licensing commitment via the link they received by email.

Please advise if there are further actions I need to take. Thanks.

rjksmith:
Rebased to main

index.bs
```diff
@@ -5544,7 +5545,8 @@ enum DirectionSetting { "" /* horizontal */, "rl", "lr" };
 enum LineAlignSetting { "start", "center", "end" };
 enum PositionAlignSetting { "line-left", "center", "line-right", "auto" };
 enum AlignSetting { "start", "center", "end", "left", "right" };
-[Exposed=Window]
+[Exposed=Window,
+ Constructor(double startTime, unrestricted double endTime, DOMString text)]
```
foolip, line 5549: Constructor an extended attribute has been removed from Web IDL. Instead of this, you'll need to add "unrestricted" below, I'll make a suggestion.

index.bs
```diff
@@ -5544,7 +5545,8 @@ enum DirectionSetting { "" /* horizontal */, "rl", "lr" };
...
 interface VTTCue : TextTrackCue {
   constructor(double startTime, double endTime, DOMString text);
```
foolip, line 5551:
```suggestion
  constructor(double startTime, unrestricted double endTime, DOMString text);
```

index.bs
```diff
@@ -5693,7 +5695,8 @@ interface VTTCue : TextTrackCue {
  interpreted as a time in seconds.</p></li>
 
  <li><p>Let |cue|'s <a>text track cue end time</a> be the value of the |endTime| argument,
- interpreted as a time in seconds.</p></li>
+ interpreted as a time in seconds or positive Infinity for an <a>unbounded text track
```
foolip, line 5698: Like in HTML, you'll need to check if the argument is -Infinity of NaN and throw a TypeError. That's because the `endTime` setter defined in HTML isn't called here, rather it's directly setting the internal "text track cue end time".

foolip (commented):
I'm not an editor of this spec, but I used to be, so some feedback. Looks good, close to the finish line now :)

index.bs
rjksmith: Thanks. Good spot. Fixed

index.bs
rjksmith: Interesting. I'd assumed that this behaviour would be inherited from `TextTrackCue.endTime` in HTML. Suitable wording added. Thanks.

index.bs
foolip: And this change here is no longer needed. Just the constructor operation below.

index.bs
foolip: Ah, so it's not the `endTime` setter that needs this, that is indeed inherited (in a sense) from `TextTrackCue`. It's the `VTTCue` constructor that needs to check for -Infinity and NaN.

index.bs
rjksmith: Changed back to `[Exposed=Window]` as requested

index.bs
rjksmith: Removed `endTime` setter wording and added equivalent check to step 3 of the `VTTCue` constructor

index.bs
```diff
@@ -5692,8 +5693,12 @@ interface VTTCue : TextTrackCue {
  <li><p>Let |cue|'s <a>text track cue start time</a> be the value of the |startTime| argument,
  interpreted as a time in seconds.</p></li>
 
- <li><p>Let |cue|'s <a>text track cue end time</a> be the value of the |endTime| argument,
- interpreted as a time in seconds.</p></li>
+ <li><p>If the value of the |endTime| argument is negative Infinity or a Not-a-Number (NaN) value,
+ then throw a <a
+ href="https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror">TypeError
+ </a> exception. Otherwise, let |cue|'s <a>text track cue end time</a> be the value of the |endTime|
+ argument, interpreted as a time in seconds or positive Infinity for an <a>unbounded text track
```
foolip, line 5700: Like in HTML, I don't think it's necessary to say "interpreted as a time in seconds or positive Infinity", instead just set the value.

foolip (approved):
Looks good, just one remaining comment from me.

I'm not sure who will review and merge this, hope you know who to poke, @rjksmith :)

index.bs
rjksmith: I agree. I was trying to match the existing wording for `startTime`, but it's unnecessary. Fixed.

foolip (approved):
This now looks done to me. Thanks for persisting, @rjksmith!

rjksmith:
Excellent. You're welcome and thanks for your guidance @foolip

rjksmith:
@gkatsev This PR is now ready for review and merge - as are whatwg/html#5953 and web-platform-tests/wpt#28394. Please co-ordinate with @foolip to synchronise these changes and let me know if there are any problems. Thanks.

himorin:
> @himorin I've signed the non-participant agreement but still see a [failing check](https://labs.w3.org/repo-manager/pr/id/w3c/webvtt/493) on this issue.

sorry on late reply. it takes some time (depends on health and delays - periodical processing of many of back end services) for IPP, and please count some time for these... (and seems fine for now on IPP / ipr.)

rjksmith:
@gkatsev whatwg/html#5953 and web-platform-tests/wpt#28394 have been merged. [As mentioned above](https://github.com/w3c/webvtt/pull/493#issuecomment-815688161), Web Platform Tests rely on VTTCue and so will fail spuriously without this PR. Please now review and merge #493 to avoid a mismatch with WPT and let me know if there are any problems. Thanks.

chrisn:
This change looks good to me, and seems straightforward to align with the TextTrackCue change in HTML. Is it OK to merge? Can the questions around the WebVTT document format be followed up separately to merging this change?

gkatsev (approved):
Thanks @rjksmith and thanks @foolip for helping out!

I believe this is good to merge and the question of updating the syntax to allow unbounded endTime has been moved to #496. It's on the agenda for this Thursday's TTWG meeting.

index.bs
```diff
@@ -54,6 +54,7 @@ urlPrefix: https://html.spec.whatwg.org/multipage/
             text: rules for extracting the chapter title
             text: text track cue start time
             text: text track cue end time
+            text: unbounded text track cue
```
foolip, line 57: Turns out this term isn't linked any longer, so this line could be removed. Or maybe a note could be added saying that a cue with `endTime` Infinity is an unbounded text track cue, linking to it.

rjksmith: Well spotted. I'll either add a note as you suggest or link to https://html.spec.whatwg.org/multipage/media.html#unbounded-text-track-cue.

chrisn:
> Thanks @rjksmith and thanks @foolip for helping out!
> 
> I believe this is good to merge and the question of updating the syntax to allow unbounded endTime has been moved to #496. It's on the agenda for this Thursday's TTWG meeting.

Thanks @gkatsev :-) I'll join the meeting tomorrow.

index.bs
rjksmith: Added note with a link to [unbounded text track cue](https://html.spec.whatwg.org/multipage/embedded-content.html#unbounded-text-track-cue), though I'm unsure whether my wording is correct as I was unable to find a comparable example.

rjksmith:
I'm happy to join too if someone can send me an invite. Thanks.

index.bs
gkatsev: Sounds like reasonable wording to me.

gkatsev (approved):

rjksmith:
@gkatsev Fantastic. Many thanks for your swift response. This enables the WPT update in web-platform-tests/wpt#28394 which allows browser implementers to validate their code for unbounded cue support, and concludes the changes required for whatwg/html#5953.

I look forward to the constructive discussion of possible changes to WebVTT in w3c/webvtt#496.

PR #507, opened 2022-11-23
Update colors of webvtt region diagram

### LINKED ISSUE #487: Figure 1: use more accessible colors
The green/orange colors in [figure 1 (§ 3.4)](https://github.com/w3c/webvtt/blob/gh-pages/webvtt-region-diagram.png) are a bit difficult to distinguish for people with red-green colour blindness (like me). E.g. the following colors could be used here, changing orange to red ([source](https://github.com/w3c/imsc/issues/391#issuecomment-397728217)):

- `#50FF50`
- `#FF5F5F`

gkatsev:
Fixes #487

I didn't have the source of the png, so, fiddled with it a bit. Might be better to redo the entire diagram as an SVG or something, though.


gkatsev:
I think this would be better served by creating a brand new SVG diagram instead.

PR #510, opened 2023-02-17
[Editorial] Add "noexport" to :past and :future dfns

tidoust:
The `:past` and `:future` pseudo-classes are formally defined in Selectors 4, which the WebVTT spec correctly references. It is fine to re-define these terms locally. The problem is that previous commit flagged these definitions as being selectors, which is correct as well but also made Bikeshed export them.

In other words, these definitions now conflict with those in Selectors 4.

This update flags these definitions with a `noexport` attribute to tell Bikeshed to keep these definitions local.

***
<a href="https://pr-preview.s3.amazonaws.com/tidoust/webvtt/pull/510.html" title="Last updated on Feb 27, 2023, 6:05 PM UTC (41d5e42)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webvtt/510/f6b3eab...tidoust:41d5e42.html" title="Last updated on Feb 27, 2023, 6:05 PM UTC (41d5e42)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5480,7 +5480,7 @@ would have been applied to the <a>list of WebVTT Node Objects</a>, must instead 
 <p>The '':past'' and '':future'' pseudo-classes sometimes match <a lt="WebVTT Node Object">WebVTT
 Node Objects</a>. [[!SELECTORS4]]</p>
 
-<p>The <dfn selector>:past</dfn> pseudo-class only matches <a lt="WebVTT Node Object">WebVTT Node Objects</a>
+<p>The <dfn selector noexport>:past</dfn> pseudo-class only matches <a lt="WebVTT Node Object">WebVTT Node Objects</a>
 that are <i>in the past</i>.</p>
 
 <p algorithm="in the past">A <a>WebVTT Node Object</a> |c| is <dfn abstract-op>in the past</dfn> if, in a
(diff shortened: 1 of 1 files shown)


gkatsev (approved):
Thanks!

I was wondering how I would go about verifying this? I tried running the CSS Selectors 4 document through bikeshed, but it didn't complain about this. Also, not seeing bikeshed complain about this when running it against webvtt.

tidoust:
> I was wondering how I would go about verifying this?

That verification is done when attempting to curate data crawled from specs in [Webref](https://github.com/w3c/webref#webref) (curated data is then fed into Bikeshed and ReSpec to populate their respective cross-references database). When a problem is detected, data curation crashes and someone needs to look into it to (1) identify the root, (2) report it, and (3) prepare a temporary patch so that curation may resume (such as the one created for WebVTT for this particular issue, at [webvtt.json.patch](https://github.com/w3c/webref/blob/main/ed/csspatches/webvtt.json.patch)). In other words, me getting back to you is essentially the only way to know about such cross-spec issues for now! Well, you are also welcome to play with Webref data yourself, of course ;)

We're trying to improve automation, e.g. through automated ["anomaly" reports](https://github.com/w3c/strudy#strudy) but it takes time... In this particular case, editing tools could perhaps check their cross-references database and alert spec authors when they export a definition that already exists elsewhere.

gkatsev:
Thanks!

Issue #485, opened 2020-06-05
Where should "headers" go relative to the `WEBVTT` magic string?


icbaker:
We're implementing a WebVTT parser that also handles the `X-TIMESTAMP-MAP` header specified by [Section 3.5 of HLS RFC 8216](https://tools.ietf.org/html/rfc8216#section-3.5).

A [question](https://github.com/google/ExoPlayer/issues/7462) has come up about where headers should be in a WebVTT file. The spec seems to only mention "headers" in [6.1 WebVTT file parsing](https://www.w3.org/TR/webvtt1/#webvtt-parser-algorithm), and by my interpretation of steps 6 - 11 they must be on lines directly beneath `WEBVTT` (i.e. no blank lines in between). It seems step 6 allows for additional text after `WEBVTT` but then ignores/skips it in step 7. Then step 9 advances to the next line and step 11 parses any headers.

It seems by this algorithm this would be a valid file:
```
WEBVTT
X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:900000

00:00:46.582 --> 00:00:48.305
  First cue
```

However [4.1 WebVTT file structure](https://www.w3.org/TR/webvtt1/#file-structure) seems to require one or more blank lines directly underneath `WEBVTT` (step 4) and doesn't mention headers at all. It does allow text immediately after `WEBVTT` on the first line (i.e. the same text that is skipped by the parsing algorithm) (step 3).

So by that interpretation, my example file above is invalid, and the only place to add 'extra' info to the start of the file is directly after `WEBVTT`, e.g.:
```
WEBVTT X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:900000

00:00:46.582 --> 00:00:48.305
  First cue
```

But the `X-TIMESTAMP-MAP` data is now in the section that's skipped by step 7 of the algorithm in 6.1 WebVTT file parsing.

So I guess there's two questions:
* Do 4.1 and 6.1 in the spec disagree about what lines can follow `WEBVTT`?
* Should headers be directly beneath `WEBVTT` or on the same line, or is either valid?

fsoder:
> * Do 4.1 and 6.1 in the spec disagree about what lines can follow `WEBVTT`?

No - one defines syntax and the other parsing. The note in [section 2.1](https://www.w3.org/TR/webvtt1/#conformance-classes) explains what the difference is.

> * Should headers be directly beneath `WEBVTT` or on the same line, or is either valid?

[...]

Issue #512, opened 2023-03-28
VTT Type Proposal for time-coded general-flash metadata 


cookiecrook:
Earlier today, Apple released a [new strobing mitigation media accessibility accommodation in tvOS 16.4, iOS 16.4, and macOS 13.3, called "Dim Flashing Lights."](https://developer.apple.com/accessibility/#whats-new) (Additional links below)

Prior to this, the current typical behavior with flashing media is that a viewer is warned *once* before before a video plays. Even in a feature length film, there is no indication of when the flashing may occur or information about the intensity and type of strobing.

On Apple devices, we can now adjust for some strobing/flashing in real time with onboard hardware and software, but as Apple's TV+ content is viewable on many others platforms, we'd also like to be able to do this on non-Apple hardware. 

We think proposing this as an accessibility standard makes sense, as other platforms may wish to adopt their own form of strobing mitigation or warnings.

So we'd like to propose using a WebVTT metadata track to allow streaming services and content owners to denote areas of flashing in video time codes similar to a caption track.

### Minimum proposal
```webvtt
WEBVTT

1
00:00:10.123 --> 00:00:15.432
{
  type: "video.strobing.general-flash",
  level: "75",
}

2
00:00:47.462 --> 00:01:04.028
{
  type: "video.strobing.general-flash",
  level: "100",
}

```

We're not particular to the proposed key names or value formats, but the type value above aligns with an expected update to HLS, and accounts for future expansion into other common strobing patterns such as `red-flash` and `spatial-pattern`.

### Variant proposal that accounts for algorithm and version

Ideally, in order to ensure all content publishers, software developers, and users have the same understanding and representation of the level, the VTT cue format could account for different, documented algorithms and versions for each of those. 

[Apple published a new general flash algorithm (PDF)](https://developer.apple.com/accessibility/downloads/Video-Flashing-Reduction-12.pdf) for use alongside [the related native platform API](https://developer.apple.com/documentation/mediaaccessibility/flashing_lights/responding_to_changes_in_the_flashing_lights_setting) or for [use in other contexts](https://github.com/apple/VideoFlashingReduction/). 

In order to differentiate from other flash pattern algorithms in WebVTT (Harding, for example), the metadata cue could contain a published version and name or URI, such as the following:

```webvtt
WEBVTT

1
00:00:10.123 --> 00:00:15.432
{
  type: "video.strobing.general-flash",
  level: "75",
  test-uri: "https://developer.apple.com/accessibilty/...",
[...]

PR #523, opened 2024-02-03
New ATTRIBUTES block

### LINKED ISSUE #511: VTT Metadata Cue format is ambiguous; some metadata may be unintentionally presented to the user in a context outside HTML
VTT Metadata Cue format is ambiguous; some metadata may be unintentionally presented to the user in a context outside HTML.

Consider clarifying that metadata cues SHOULD or MUST be JSON-formatted, or at least be unambiguously formatted, such as starting with `{` and ending with `}\n\n` to retain backwards compatibility with the JSON usage documented in the VTT spec.

```webvtt
WEBVTT

1
00:00:10.123 --> 00:00:15.432
{
  key: "value"
}
```

### Background

[§ 4.2.1. WebVTT metadata text](https://w3c.github.io/webvtt/#metadata-text) (Normative) defines metadata text as:

> WebVTT metadata text consists of any sequence of zero or more characters other than U+000A LINE FEED (LF) characters and U+000D CARRIAGE RETURN (CR) characters, each optionally separated from the next by a WebVTT line terminator. (In other words, any text that does not have two consecutive WebVTT line terminators and does not start or end with a WebVTT line terminator.)
> 
> WebVTT metadata text cues are only useful for scripted applications (e.g. using the metadata text track kind in a HTML text track).

[§ 1.7. Metadata example](https://w3c.github.io/webvtt/#introduction-metadata) (Informative) clarifies: 

> A WebVTT file can consist of time-aligned metadata.
> 
> Metadata can be *any string* and is *often provided as a JSON construct*.

### Problem 

*"Metadata can be any string"* results in a format that is ambiguous, and therefore may be presented to the user unintentionally.

In an HTML `<video>` element, this ambiguity is resolved by the author providing a `kind="metadata"` attribute on the text track.

But there isn't a logical place to duplicate this disambiguation in some other VTT contexts, including when they are embedded in some media container formats.

### Proposed Solution

Consider clarifying that metadata cues SHOULD or MUST be JSON-formatted, or at least be unambiguously formatted, such as starting with `{` and ending with `}\n\n` to retain backwards compatibility with the JSON usage documented in the VTT spec.

Additional context for this change request coming in another issue.

- Todo: Link new VTT Type Proposal for Strobe Mitigation

cookiecrook:
Closes #511

Add new ATTRIBUTES block parsing rules and examples as discussed in #511.

This is also a prerequisite for:
- https://github.com/w3c/webvtt/issues/512

…and its duplicate in the WHATWG repo:
https://github.com/whatwg/html/issues/11333

***
<a href="https://pr-preview.s3.amazonaws.com/cookiecrook/webvtt/pull/523.html" title="Last updated on Mar 9, 2026, 7:24 PM UTC (b60d4b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webvtt/523/bd70b0a...cookiecrook:b60d4b9.html" title="Last updated on Mar 9, 2026, 7:24 PM UTC (b60d4b9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -573,7 +573,7 @@ WEBVTT
 
 ATTRIBUTES
 kind: subtitles
-srclang: es-mx
+lang: es-mx
 label: Español
 
 NOTE
@@ -597,7 +597,7 @@ WEBVTT
 
 ATTRIBUTES
 kind: captions
-srclang: es-mx
+lang: es-mx
 label: Español (SDH)
 
 NOTE
@@ -718,13 +718,13 @@ this case, "music".
 
 <div class="example">
 
- <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
+ <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent audible or braille descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
  <pre>
 WEBVTT
 
 ATTRIBUTES
 kind: descriptions
-srclang: en-us
+lang: en-us
 label: English (AD)
 
 NOTE
@@ -798,11 +798,6 @@ WEBVTT
 ATTRIBUTES
 kind: metadata
 
-NOTE
-The Timed Text Working Group is discussing a registry for metadata `type` 
-values, such as `type: video-thumbnails` or `type: video-flash-avoidance`. 
-See webvtt issues #511 and #512 for more info.
-
 00:00:01.959 --> 00:00:02.938
 {
  "src": "https://cdn.example.com/thumbnails.jpg#xywh=0,0,284,160",
@@ -812,6 +807,11 @@ See webvtt issues #511 and #512 for more info.
  }
 }
  </pre>
+ 
+ <p class="note">The Timed Text Working Group is discussing a registry for metadata <code>type</code> 
+values, such as <code>type: video-thumbnails</code> or <code>type: video-flash-avoidance</code>. 
+See WebVTT issues <a href="https://github.com/w3c/webvtt/issues/511">#511</a> and <a href="https://github.com/w3c/webvtt/issues/512">#512</a> for more info.</p>
+
 
 </div>
 
@@ -4294,7 +4294,7 @@ follows:</p>
    <dt>If the key name is "<code>kind</code>"</dt>
    <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
 
-   <dt>If the key name is "<code>srclang</code>"</dt>
+   <dt>If the key name is "<code>lang</code>"</dt>
    <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang">the srclang attribute</a> of a track element according to the HTML Standard.</dd>
 
    <dt>If the key name is "<code>label</code>"</dt>
--- a/index.bs
+++ b/index.bs
@@ -563,6 +563,62 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+
+
+<div class="example">
+
+ <p>In this example, an optional WebVTT attributes object is used to define the source language and its label in a subtitle/caption selection menu.</p>
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: subtitles
+srclang: es-mx
+label: Español
+
+NOTE
+Standard subtitles (unlike CC or SDH captions) typically 
+translate spoken dialog or signage, but not audible sounds 
+effects like "dogs barking."
+
+1
+00:00:10.123 --> 00:00:15.432
+¡Hola! ¿Qué tál?
+ </pre>
+
+</div>
+
+
+<div class="example">
+
+ <p>In this example, an optional WebVTT attributes object is used to differentiate captions from standard subtitles.</p>
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: captions
+srclang: es-mx
+label: Español (SDH)
+
+NOTE
+Captions (SDH aka Subtitles for the Deaf and Hard-of-Hearing) 
+typically include spoken dialog as well as important audible 
+sounds such as "floor boards creak", "dogs barking", or in 
+this case, "music".
+
+1
+00:00:10.123 --> 00:00:15.432
+¡Hola! ¿Qué tál?
+
+2
+00:00:47.462 --> 00:01:04.028
+[♫ música ♫]
+ </pre>
+
+</div>
+
+
+
 <h3 id=introduction-comments>Comments in WebVTT</h3>
 
 <p><i>This section is non-normative.</i></p>
@@ -658,6 +714,32 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+
+
+<div class="example">
+
+ <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent video descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: descriptions
+srclang: en-us
+label: English (AD)
+
+NOTE
+VTT-based descriptions are meant to render as text-to-speech audio or braille,
+for blind or deafblind audiences, not usually as visual captions on screen. 
+As such, the option/label might be displayed in an audio menu or elsewhere. 
+
+1
+00:00:10.123 --> 00:00:15.432
+A young girl tiptoes down a dark hallway.
+ </pre>
+
+</div>
+
+
 <h3 id=introduction-metadata>Metadata example</h3>
 
 <p><i>This section is non-normative.</i></p>
@@ -671,11 +753,14 @@ signifies the end of the WebVTT cue.</p>
 
 <div class="example">
 
- <p>In this example, a talk is split into each slide being a chapter.</p>
+ <p>In this example, topics mentioned in a talk are provided as URLs for reference.</p>
 
  <pre>
  WEBVTT
 
+ ATTRIBUTES
+ kind: metadata
+ 
  NOTE
  Thanks to http://output.jsbin.com/mugibo
 
@@ -700,6 +785,28 @@ signifies the end of the WebVTT cue.</p>
   "lat" : "36.198269",
   "long": "137.2315355"
  }
+ </pre>
+
+</div>
+
+<div class="example">
+
+ <p>In this example, a sequence of video thumbnails and their text alternative are made available for the playback UI.</p>
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: metadata
+type: video-thumbnails
+
+00:00:01.959 --> 00:00:02.938
+{
+ "src": "https://cdn.example.com/thumbnails.jpg#xywh=0,0,284,160",
+ "alt": {
+  "en-us": "Miguel crosses the marigold bridge to the land of the dead.",
+  "es-mx": "Miguel cruza el puente marigold hacia la tierra de los muertos."
+ }
+}
  </pre>
 
 </div>
@@ -1650,6 +1757,28 @@ SIGN).</p>
 
 <p>When interpreted as a number, a <a>WebVTT percentage</a> must be in the range 0..100.</p>
 
+<p>A <dfn>WebVTT attributes object</dfn> consists of the following components, in the given order:</p>
+<ol>
+ <li>The string "<code>ATTRIBUTES</code>".</li>
+ <li>
+  The following components, in the given order:
+  <ol>
+   <li>A <a>WebVTT line terminator</a>.</li>
+   <li>Zero or more key/value pairs, parsed in the given order:
+    <ol>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of any sequence of one or more (TODO: /A-Za-z/?) characters.</li>
+     <li>A single U+003A COLON character ("<code>:</code>").</li>
+     <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
+     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than U+000A LINE FEED (LF) characters and U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
+     <li>A <a>WebVTT line terminator</a>.</li>
+    </ol>
+   </li>
+   <li>A final <a>WebVTT line terminator</a> to complete the WebVTT attributes object.</li>
+  </ol>
+ </li>
+</ol>
+
+
 <p>A <dfn>WebVTT comment block</dfn> consists of the following components, in the given order:</p>
 
 <ol>
@@ -1687,7 +1816,7 @@ separated from the next by a <a>WebVTT line terminator</a>. (In other words, any
 have two consecutive <a lt="WebVTT line terminator">WebVTT line terminators</a> and does not start
 or end with a <a>WebVTT line terminator</a>.)</p>
 
-<p><a>WebVTT metadata text</a> cues are only useful for scripted applications (e.g. using the
+<p><a>WebVTT metadata text</a> cues were originally intended for scripted applications (e.g. using the
 <code>metadata</code> <a>text track kind</a> in a HTML <a>text track</a>).</p>
 
 
--- a/index.bs
+++ b/index.bs
@@ -578,7 +578,7 @@ label: Español
 
 NOTE
 Standard subtitles (unlike CC or SDH captions) typically 
-translate spoken dialog or signage, but not audible sounds 
+translate spoken dialog or signage, but not audible sound 
 effects like "dogs barking."
 
 1
@@ -1773,25 +1773,21 @@ SIGN).</p>
    <li>A <a>WebVTT line terminator</a>.</li>
    <li>Zero or more key/value pairs, parsed in the given order:
     <ol>
-     <li>A <dfn>WebVTT attribute key</dfn> consisting of <code>[A-Za-z_][0-9A_Za-z_]*</code>:
-      <ul>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of: (<code>[A-Za-z_][0-9A_Za-z_]*</code>)
+      <ol>
        <li>Any one of the following:
         <ul>
-         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
-         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
-         <li>U+005F LOW LINE _ ("underscore")</li>
+         <li>Any <a href="https://infra.spec.whatwg.org/#ascii-alpha">ASCII Alpha</a> character</li>
+         <li>U+005F LOW LINE ("_" underscore)</li>
         </ul>
        </li>
        <li>Optionally followed by zero or more of the following:
         <ul>
-         <li>U+0030 DIGIT ZERO ("0") through U+0039 DIGIT NINE ("9")</li>
-         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
-         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
+         <li>Any <a href="https://infra.spec.whatwg.org/#ascii-alphanumeric">ASCII Alphanumeric</a> character</li>
          <li>U+005F LOW LINE ("_" underscore)</li>
         </ul>
        </li>
-       <li class="ednote">Editorial Note: Should this `key` token range be an external reference to the character range for HTML TagName or ECMAScript variables? If so, which reference?</li>
-      </ul>
+      </ol>
      </li>
      <li>A single U+003A COLON character ("<code>:</code>").</li>
      <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
@@ -4319,9 +4315,10 @@ follows:</p>
 <ol algorithm="WebVTT type attribute attribute parsing">
  <li>TODO: This could reference a new TBD W3C Note or Evergreen list of acknowledged kind subtypes, along with a reference to the specification for each, which clarify the usage or define further parsing rules of each type. For example:
   <ul>
-   <li>metadata subtype: time-coded video poster thumbnails (common de facto use for scrubbing but no spec)</li>
+   <li>metadata subtype: time-coded video poster thumbnails (<a href="https://github.com/w3c/webvtt/issues/511#issuecomment-1487497644">common de facto use for preview scrubbing</a> but no spec exists)</li>
    <li>metadata subtype: <a href="https://github.com/w3c/webvtt/issues/512">WebVTT Issue 512: time-coded flash metadata</a></li>
-   <li>caption or description subtype: text equivalent of audio description audio track (used for braille displays)</li>
+   <li>metadata subtype: <a href="https://github.com/w3ctag/design-reviews/issues/952">video chapter times, thumbnails, and titles</a>
+   <li>caption or description subtype: text equivalent of audio description dialogue audio track (<a href="https://www.w3.org/2022/09/14-audio-descriptions-minutes.html">VTT descriptions rendered on braille displays</a> used by blind and deafblind individuals)</li>
    <li>etc.</li>
   </ul>
  </li>
--- a/index.bs
+++ b/index.bs
@@ -1585,7 +1585,7 @@ with the <a>MIME type</a> <code>text/vtt</code>. [[!RFC3629]]</p>
  <li>Two or more <a lt="WebVTT line terminator">WebVTT line terminators</a> to terminate the line
  with the file magic and separate it from the rest of the body.</li>
 
- <li>Zero or one <a lt="WebVTT attributes object">WebVTT attributes object</a> followed by one or 
+ <li>Zero or one <a lt="WebVTT attributes block">WebVTT attributes block</a> followed by one or 
  more <a lt="WebVTT line terminator">WebVTT line terminators</a>.</li>
 
  <li>Zero or more <a lt="WebVTT region definition block">WebVTT region definition blocks</a>, <a
@@ -1764,7 +1764,7 @@ SIGN).</p>
 
 <p>When interpreted as a number, a <a>WebVTT percentage</a> must be in the range 0..100.</p>
 
-<p>A <dfn>WebVTT attributes object</dfn> consists of the following components, in the given order:</p>
+<p>A <dfn>WebVTT attributes block</dfn> consists of the following components, in the given order:</p>
 <ol>
  <li>The string "<code>ATTRIBUTES</code>".</li>
  <li>
@@ -1773,17 +1773,38 @@ SIGN).</p>
    <li>A <a>WebVTT line terminator</a>.</li>
    <li>Zero or more key/value pairs, parsed in the given order:
     <ol>
-     <li>A <dfn>WebVTT attribute key</dfn> consisting of (TODO: ref HTML tagname or ECMAScript variable char ranges? or identifier regex <code>[a-zA-Z_][0-9a-zA-Z_]*</code>).</li>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of <code>[A-Za-z_][0-9A_Za-z_]*</code>:
+      <ul>
+       <li>Any one of the following:
+        <ul>
+         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
+         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
+         <li>U+005F LOW LINE _ ("underscore")</li>
+        </ul>
+       </li>
+       <li>Optionally followed by zero or more of the following:
+        <ul>
+         <li>U+0030 DIGIT ZERO ("0") through U+0039 DIGIT NINE ("9")</li>
+         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
+         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
+         <li>U+005F LOW LINE ("_" underscore)</li>
+        </ul>
+       </li>
+       <li class="ednote">Editorial Note: Should this `key` token range be an external reference to the character range for HTML TagName or ECMAScript variables? If so, which reference?</li>
+      </ul>
+     </li>
      <li>A single U+003A COLON character ("<code>:</code>").</li>
      <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
      <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
      <li>A <a>WebVTT line terminator</a>.</li>
     </ol>
    </li>
-   <li>A final <a>WebVTT line terminator</a> to complete the WebVTT attributes object.</li>
+   <li>A final <a>WebVTT line terminator</a> to complete the WebVTT attributes block.</li>
   </ol>
  </li>
 </ol>
+
+<p>Process the <a>WebVTT attributes block</a> key/value pairs according to the <a>WebVTT attributes key/value parsing rules</a>.</p>
 
 
 <p>A <dfn>WebVTT comment block</dfn> consists of the following components, in the given order:</p>
@@ -4263,6 +4284,47 @@ follows:</p>
  pre-order, depth-first traversal, excluding <a lt="WebVTT Ruby Text Object">WebVTT Ruby Text
  Objects</a> and their descendants.</p></li>
 
+</ol>
+
+
+<p>The <dfn>WebVTT attributes key/value parsing rules</dfn> consist of the following algorithm.</p>
+
+<ol algorithm="WebVTT attributes block parsing">
+ <li>Let |input| be the list of key/value pairs from a <a>WebVTT attributes block</a>.</li>
+ <li>
+  How the attribute is processed depends on its key name, as follows:
+  <dl>
+
+   <dt>If the key name is "<code>kind</code>"</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>srclang</code>"</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang">the srclang attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>label</code>"</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-label">the label attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>type</code>" (TODO: For clarity, should this be "subkind" or "kind_subtype" instead?)</dt>
+   <dd>Process the value according to the <a>WebVTT type attribute parsing rules</a>.
+
+   <dt>Otherwise</dt>
+   <dd>Ignore the key/value pair.</dd>
+
+  </dl>
+ </li>
+</ol>
+
+<p>The <dfn>WebVTT type attribute parsing rules</dfn> consist of the following algorithm.</p>
+
+<ol algorithm="WebVTT type attribute attribute parsing">
+ <li>TODO: This could reference a new TBD W3C Note or Evergreen list of acknowledged kind subtypes, along with a reference to the specification for each, which clarify the usage or define further parsing rules of each type. For example:
+  <ul>
+   <li>metadata subtype: time-coded video poster thumbnails (common de facto use for scrubbing but no spec)</li>
+   <li>metadata subtype: <a href="https://github.com/w3c/webvtt/issues/512">WebVTT Issue 512: time-coded flash metadata</a></li>
+   <li>caption or description subtype: text equivalent of audio description audio track (used for braille displays)</li>
+   <li>etc.</li>
+  </ul>
+ </li>
 </ol>
 
 
--- a/index.bs
+++ b/index.bs
@@ -797,7 +797,11 @@ WEBVTT
 
 ATTRIBUTES
 kind: metadata
-type: video-thumbnails
+
+NOTE
+The Timed Text Working Group is discussing a registry for metadata `type` 
+values, such as `video-thumbnails` or `video-flash-avoidance`. 
+See webvtt issues #511 and #512 for more info.
 
 00:00:01.959 --> 00:00:02.938
 {
--- a/index.bs
+++ b/index.bs
@@ -4300,33 +4300,12 @@ follows:</p>
    <dt>If the key name is "<code>label</code>"</dt>
    <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-label">the label attribute</a> of a track element according to the HTML Standard.</dd>
 
-<!-- See https://github.com/w3c/webvtt/issues/525
-   <dt>If the key name is "<code>type</code>" (TODO: For clarity, should this be "subkind" or "kind_subtype" instead?)</dt>
-   <dd>Process the value according to the <a>WebVTT type attribute parsing rules</a>.
--->
-
    <dt>Otherwise</dt>
    <dd>Ignore the key/value pair.</dd>
 
   </dl>
  </li>
 </ol>
-
-<!-- See https://github.com/w3c/webvtt/issues/525
-<p>The <dfn>WebVTT type attribute parsing rules</dfn> consist of the following algorithm.</p>
-
-<ol algorithm="WebVTT type attribute attribute parsing">
- <li>TODO: This could reference a new TBD W3C Note or Evergreen list of acknowledged kind subtypes, along with a reference to the specification for each, which clarify the usage or define further parsing rules of each type. For example:
-  <ul>
-   <li>metadata subtype: time-coded video poster thumbnails (<a href="https://github.com/w3c/webvtt/issues/511#issuecomment-1487497644">common de facto use for preview scrubbing</a> but no spec exists)</li>
-   <li>metadata subtype: <a href="https://github.com/w3c/webvtt/issues/512">WebVTT Issue 512: time-coded flash metadata</a></li>
-   <li>metadata subtype: <a href="https://github.com/w3ctag/design-reviews/issues/952">video chapter times, thumbnails, and titles</a>
-   <li>caption or description subtype: text equivalent of audio description dialogue audio track (<a href="https://www.w3.org/2022/09/14-audio-descriptions-minutes.html">VTT descriptions rendered on braille displays</a> used by blind and deafblind individuals)</li>
-   <li>etc.</li>
-  </ul>
- </li>
-</ol>
--->
 
 <h2 id=rendering>Rendering</h2>
 
--- a/index.bs
+++ b/index.bs
@@ -800,7 +800,7 @@ kind: metadata
 
 NOTE
 The Timed Text Working Group is discussing a registry for metadata `type` 
-values, such as `video-thumbnails` or `video-flash-avoidance`. 
+values, such as `type: video-thumbnails` or `type: video-flash-avoidance`. 
 See webvtt issues #511 and #512 for more info.
 
 00:00:01.959 --> 00:00:02.938
@@ -1773,7 +1773,7 @@ SIGN).</p>
    <li>A <a>WebVTT line terminator</a>.</li>
    <li>Zero or more key/value pairs, parsed in the given order:
     <ol>
-     <li>A <dfn>WebVTT attribute key</dfn> consisting of any sequence of one or more (TODO: ref HTML tagname char or ECMAScript variable char ranges) characters.</li>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of (TODO: ref HTML tagname or ECMAScript variable char ranges? or identifier regex <code>[a-zA-Z_][0-9a-zA-Z_]*</code>).</li>
      <li>A single U+003A COLON character ("<code>:</code>").</li>
      <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
      <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
--- a/index.bs
+++ b/index.bs
@@ -718,7 +718,7 @@ this case, "music".
 
 <div class="example">
 
- <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent video descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
+ <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
  <pre>
 WEBVTT
 
--- a/index.bs
+++ b/index.bs
@@ -1766,10 +1766,10 @@ SIGN).</p>
    <li>A <a>WebVTT line terminator</a>.</li>
    <li>Zero or more key/value pairs, parsed in the given order:
     <ol>
-     <li>A <dfn>WebVTT attribute key</dfn> consisting of any sequence of one or more (TODO: /A-Za-z/?) characters.</li>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of any sequence of one or more (TODO: ref HTML tagname char or ECMAScript variable char ranges) characters.</li>
      <li>A single U+003A COLON character ("<code>:</code>").</li>
      <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
-     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than U+000A LINE FEED (LF) characters and U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
+     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
      <li>A <a>WebVTT line terminator</a>.</li>
     </ol>
    </li>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -563,6 +563,62 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+2
+00:00:47.462 --> 00:01:04.028
+[♫ música ♫]
+ </pre>
+
+</div>
+
+
+
 <h3 id=introduction-comments>Comments in WebVTT</h3>
```
cookiecrook, line 622: I didn't change this line, and I'm not sure what the dark red highlighting means.

index.bs
```diff
@@ -1650,6 +1757,28 @@ SIGN).</p>
 
 <p>When interpreted as a number, a <a>WebVTT percentage</a> must be in the range 0..100.</p>
```
cookiecrook, line 1759: Hoping for validation on this parsing for now. More coming later.

index.bs
```diff
@@ -1650,6 +1757,28 @@ SIGN).</p>
...
+ <li>
+  The following components, in the given order:
+  <ol>
+   <li>A <a>WebVTT line terminator</a>.</li>
+   <li>Zero or more key/value pairs, parsed in the given order:
+    <ol>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of any sequence of one or more (TODO: /A-Za-z/?) characters.</li>
+     <li>A single U+003A COLON character ("<code>:</code>").</li>
+     <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
+     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than U+000A LINE FEED (LF) characters and U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
```
cookiecrook, line 1772:
```suggestion
     <li>A <dfn>WebVTT attribute key</dfn> consisting of any sequence of one or more (TODO: ref HTML tagname char or ECMAScript variable char ranges) characters.</li>
     <li>A single U+003A COLON character ("<code>:</code>").</li>
     <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
```

index.bs
```diff
@@ -704,6 +789,28 @@ signifies the end of the WebVTT cue.</p>
...
 
+<div class="example">
+
+ <p>In this example, a sequence of video thumbnails and their text alternative are made available for the playback UI.</p>
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: metadata
+type: video-thumbnails
```
cookiecrook, line 800: @eric-carlson maybe this would be ready to merge if I removed this undefined `type` line for now. There is more to be done with `type` but we can update this example to re-include type once more specifics have been worked out for the type registry. What do you think?
```suggestion
```

index.bs
cookiecrook: Eric mentioned he has a prototype of this now.

index.bs
cookiecrook:
```suggestion

NOTE
The Timed Text Working Group is discussing a type registry for metadata 
`type` values, such as `video-thumbnails` or `video-flash-avoidance`. 
See webvtt issues #511 and #512 for more info.
```

cookiecrook:
.

index.bs
```diff
@@ -4130,6 +4287,47 @@ follows:</p>
...
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>srclang</code>"</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang">the srclang attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>label</code>"</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-label">the label attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>type</code>" (TODO: For clarity, should this be "subkind" or "kind_subtype" instead?)</dt>
+   <dd>Process the value according to the <a>WebVTT type attribute parsing rules</a>.
```
cookiecrook, line 4308: This value (and the following 'type attribute processin' algo) will probably be commented for the PR tied to issue #511 and added in later with #512 or otherwise. Including here to get further comments from other VTT folks.

index.bs
```diff
@@ -4130,6 +4287,47 @@ follows:</p>
...
+<ol algorithm="WebVTT type attribute attribute parsing">
+ <li>TODO: This could reference a new TBD W3C Note or Evergreen list of acknowledged kind subtypes, along with a reference to the specification for each, which clarify the usage or define further parsing rules of each type. For example:
+  <ul>
+   <li>metadata subtype: time-coded video poster thumbnails (common de facto use for scrubbing but no spec)</li>
+   <li>metadata subtype: <a href="https://github.com/w3c/webvtt/issues/512">WebVTT Issue 512: time-coded flash metadata</a></li>
+   <li>caption or description subtype: text equivalent of audio description audio track (used for braille displays)</li>
+   <li>etc.</li>
+  </ul>
+ </li>
+</ol>
```
cookiecrook, line 4328: This whole section is effectively a TODO that is not required for #511 but shows the potential future direction of clarifying the `type` of metadata or other kinds of VTT. Feedback requested.

cookiecrook:
Marking as Ready for Review despite there being some TODOs remaining in the diff. Feedback requested from @gkatsev, @silviapfeiffer, @nigelmegitt, @eric-carlson, et al.

index.bs
```diff
@@ -1650,6 +1764,49 @@ SIGN).</p>
...
+       </li>
+       <li>Optionally followed by zero or more of the following:
+        <ul>
+         <li>U+0030 DIGIT ZERO ("0") through U+0039 DIGIT NINE ("9")</li>
+         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
+         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
+         <li>U+005F LOW LINE ("_" underscore)</li>
+        </ul>
+       </li>
+       <li class="ednote">Editorial Note: Should this `key` token range be an external reference to the character range for HTML TagName or ECMAScript variables? If so, which reference?</li>
```
cookiecrook, line 1793: 👆

index.bs
cookiecrook: Another instance of this disambiguation need outlined in:
- https://github.com/w3ctag/design-reviews/issues/952

index.bs
```diff
@@ -1650,6 +1764,49 @@ SIGN).</p>
...
+  <ol>
+   <li>A <a>WebVTT line terminator</a>.</li>
+   <li>Zero or more key/value pairs, parsed in the given order:
+    <ol>
+     <li>A <dfn>WebVTT attribute key</dfn> consisting of <code>[A-Za-z_][0-9A_Za-z_]*</code>:
+      <ul>
+       <li>Any one of the following:
+        <ul>
+         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
+         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
```
annevk, line 1781: I would reference https://infra.spec.whatwg.org/#ascii-alpha instead of these two bullet points.

index.bs
```diff
@@ -1650,6 +1764,49 @@ SIGN).</p>
...
+         <li>U+005F LOW LINE _ ("underscore")</li>
+        </ul>
+       </li>
+       <li>Optionally followed by zero or more of the following:
+        <ul>
+         <li>U+0030 DIGIT ZERO ("0") through U+0039 DIGIT NINE ("9")</li>
+         <li>U+0041 LATIN CAPITAL LETTER A through U+005A LATIN CAPITAL LETTER Z</li>
+         <li>U+0061 LATIN CAPITAL SMALL A through U+007A LATIN SMALL LETTER A</li>
```
annevk, line 1789: I would reference https://infra.spec.whatwg.org/#ascii-alphanumeric instead of these three bullet points.

cookiecrook: done

index.bs
cookiecrook: done

cookiecrook:
I think this PR is merge-ready now once a TTWG reviewer approves. The previously included TODO comments are now removed from this PR in favor of addressing that follow-on work in:
- #525

nigelmegitt (dismissed):
I think this would benefit from more clarity (i.e. more text) about the intended use of ATTRIBUTES and its relationship to attributes on the HTML `<track>` element, and any other potential usage context.

index.bs
```diff
@@ -563,6 +563,62 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+
+
+<div class="example">
+
+ <p>In this example, an optional WebVTT attributes object is used to define the source language and its label in a subtitle/caption selection menu.</p>
```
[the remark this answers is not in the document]
cookiecrook: To align with HTML [srclang](https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang).

index.bs
```diff
@@ -563,6 +563,62 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: subtitles
+srclang: es-mx
```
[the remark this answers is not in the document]
cookiecrook: All the other attributes (kind, label) are a 1:1 mirror of the track element attributes. I think [srclang](https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang) should also align.

But it's not the hill I'd die on. Open to change if others agree.

index.bs
```diff
@@ -563,6 +563,62 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+00:00:10.123 --> 00:00:15.432
+¡Hola! ¿Qué tál?
+ </pre>
+
+</div>
+
+
+<div class="example">
+
+ <p>In this example, an optional WebVTT attributes object is used to differentiate captions from standard subtitles.</p>
```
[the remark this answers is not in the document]
cookiecrook: I'm not sure I understand... Are you suggesting the value parsing algo cross reference the HTML Infra values in `track[kind]`?

index.bs
```diff
@@ -658,6 +714,32 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+
+
+<div class="example">
+
+ <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent video descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
```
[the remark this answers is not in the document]
cookiecrook:
```suggestion
 <p>In this example, a WebVTT attributes object is used to indicate the text track cues represent descriptions for the blind. Unlike subtitles or captions, these are not intended to be rendered visually.</p>
```

index.bs
```diff
@@ -563,6 +563,62 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+
+ATTRIBUTES
+kind: subtitles
+srclang: es-mx
+label: Español
+
+NOTE
+Standard subtitles (unlike CC or SDH captions) typically 
+translate spoken dialog or signage, but not audible sounds 
+effects like "dogs barking."
```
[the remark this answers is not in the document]
cookiecrook: > If there's an intended linkage between the kind attribute and the HTML [track element's kind](https://html.spec.whatwg.org/multipage/media.html#dom-track-kind) attribute then that should be explicit in the spec text.

I included that in the "WebVTT attributes key/value parsing rules." 

```html
   <dt>If the key name is "<code>kind</code>"</dt>
   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
```

Did you miss the above text, or is there some additional place where you'd like it added? Thanks for clarifying.

index.bs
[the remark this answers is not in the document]
cookiecrook: updated to use lang. see [other thread](https://github.com/w3c/webvtt/pull/523/files#r1639953968).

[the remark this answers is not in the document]
cookiecrook: updated to use lang

index.bs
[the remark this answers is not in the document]
cookiecrook: Oh, upon re-reading, it sounds like you're suggesting the examples be removed. 

FWIW, as a reader of the spec (this was my first PR to VTT), I found the pre-existing examples helpful. Perhaps others will too? The ATTRIBUTES block is defined here, so seeing how it's used with different key/value pairs seems useful to me, even if those values are defined by HTML.

index.bs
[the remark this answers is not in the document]
cookiecrook: this is changed now.

index.bs
```diff
@@ -704,6 +789,32 @@ signifies the end of the WebVTT cue.</p>
...
+ <pre>
+WEBVTT
+
+ATTRIBUTES
+kind: metadata
+
+NOTE
+The Timed Text Working Group is discussing a registry for metadata `type` 
+values, such as `type: video-thumbnails` or `type: video-flash-avoidance`. 
+See webvtt issues #511 and #512 for more info.
```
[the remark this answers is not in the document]
cookiecrook: done

index.bs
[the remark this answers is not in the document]
cookiecrook: There is one of each:
- subtitle example
- caption example
- description example
- metadata example

I'd prefer to leave all 4 examples, but perhaps we can compromise on just removing the subtitle example?

index.bs
```diff
@@ -1650,6 +1764,45 @@ SIGN).</p>
 
 <p>When interpreted as a number, a <a>WebVTT percentage</a> must be in the range 0..100.</p>
 
+<p>A <dfn>WebVTT attributes block</dfn> consists of the following components, in the given order:</p>
+<ol>
+ <li>The string "<code>ATTRIBUTES</code>".</li>
+ <li>
+  The following components, in the given order:
+  <ol>
+   <li>A <a>WebVTT line terminator</a>.</li>
```
himorin, line 1773: I am not sure, but this item seems to be out from this `<ol>`, since it divides the `ATTRIBUTES` line but not in key/value pairs

index.bs
```diff
@@ -1650,6 +1764,45 @@ SIGN).</p>
...
+       </li>
+      </ol>
+     </li>
+     <li>A single U+003A COLON character ("<code>:</code>").</li>
+     <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
+     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
+     <li>A <a>WebVTT line terminator</a>.</li>
+    </ol>
+   </li>
+   <li>A final <a>WebVTT line terminator</a> to complete the WebVTT attributes block.</li>
```
himorin, line 1798: comparing to others, do we need this additional line terminator here?

index.bs
cookiecrook: I intended that as the EOL char on the same line as ATTRIBUTES

```
ATTRIBUTES
          ^
```

But I'd be happy to combine it with the previous line, or specify some other way at the editors' direction.

index.bs
cookiecrook: That's for the extra EOL char on a blank line that ends the entire ATTRIBUTES block.

```
ATTRIBUTES
key1: value1
key2: value2

^
```

cookiecrook: Again, I'd be happy to specify that in another way at the direction of the editors.

index.bs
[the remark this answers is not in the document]
nigelmegitt: I may have missed it, but I can see why: that's appearing as part of §6.6 "WebVTT rules for extracting the chapter title" - shouldn't it be in its own numbered section?

Aside from that, what does "process the value as ..." mean? Does it mean that it takes precedence over the same attribute on the `<track>` element? Seems like we need to say something about precedence in this scenario.

index.bs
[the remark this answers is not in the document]
nigelmegitt: I still think there are too many but it's not a major complaint; as you say, many folk find examples an easy way into understanding the specification.

Also I just noticed that Example 14 is appearing in the Chapters section, but its about descriptions, so it seems to be misplaced.

nigelmegitt:
Needs a review from the Editor, @gkatsev .

nigelmegitt:
See also another operational practice described for doing this at https://github.com/w3c/webvtt/issues/485#issuecomment-2362201249 in which YouTube is doing it in the Header section. We should probably make sure that whatever we specify here aligns with practice, or if not, that folk using that alternative approach are happy to change.

index.bs
```diff
@@ -4130,6 +4283,30 @@ follows:</p>
 </ol>
 
 
+<p>The <dfn>WebVTT attributes key/value parsing rules</dfn> consist of the following algorithm.</p>
```
gkatsev, line 4286:
```suggestion
<h3 id=rules-for-parsing-attr-key-values algorithm>WebVTT Attributes key/value parsing rules</h3>

<p>The <dfn>WebVTT attributes key/value parsing rules</dfn> consist of the following algorithm.</p>
```

index.bs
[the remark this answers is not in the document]
gkatsev: I tend to like a lot of examples.

Given how similar the subtitles and captions examples are, maybe we only keep one?

gkatsev (commented):
Thanks for the work on this @cookiecrook and for your patience!

index.bs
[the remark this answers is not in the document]
gkatsev: > I may have missed it, but I can see why: that's appearing as part of §6.6 "WebVTT rules for extracting the chapter title" - shouldn't it be in its own numbered section?

I agree, it should be a separate section.

> Aside from that, what does "process the value as ..." mean? Does it mean that it takes precedence over the same attribute on the `<track>` element? Seems like we need to say something about precedence in this scenario.

I was wondering the same. Are these purely metadata or do they interact with the HTML/DOM in any way? And what is the order of precedence if both values are used? The track element wins? That might need a change in HTML?

gkatsev:
> See also another operational practice described for doing this at [#485 (comment)](https://github.com/w3c/webvtt/issues/485#issuecomment-2362201249) in which YouTube is doing it in the Header section. We should probably make sure that whatever we specify here aligns with practice, or if not, that folk using that alternative approach are happy to change.

What YouTube has definitely seems to align with it. Would be great if we can find someone from there to comment on this, if they could potentially move to the new block when it exists.

> Update: @gkatsev had the action to draft a PR to resolve #485, and that looks to me as though it would overlap strongly with this PR.

I think ultimately, particularly, with HLS's usage of the header, we do still want the header as "point of extensibility" there. But, we may want to discourage new uses of it, since the majority of these use cases could use the ATTRIBUTES block. As an example, the ECMAScript spec added the `__proto__` to the spec to get it documented but marked it as legacy: https://262.ecma-international.org/15.0/index.html#sec-object.prototype.__proto__ (I'm not aware of something specific like that in HTML or w3c specs, but I'm also not sure that there's anything against doing so).

cookiecrook:
@gkatsev wrote:
> What YouTube has definitely seems to align with it. Would be great if we can find someone from there to comment on this, if they could potentially move to the new block when it exists.

Yes, I have taken that as an action.

index.bs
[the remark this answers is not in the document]
cookiecrook: Section mismatch was probably due to my unfamiliarty with Bikeshed. I will correct it and look at your other question at the same time.

cookiecrook:
I've also taken an action to file an issue to HTML to propose any reflected attributes from the VTT to HTML Track content attributes.

cookiecrook:
From @eric-carlson:

> I just noticed all of the [VTT files in the Dash CEA608 parser test suite](https://github.com/Dash-Industry-Forum/cea608.js/tree/b40a59b1d2a2bd80df3305dc8afc70b988b218d2/test/data/vtt) have metadata in the header section, although they use = as the delimiter between key and value

cookiecrook:
**Update Summary:** After Ken Harrenstien's retirement from Google, we tried unsuccessfully several times over the last 4 months to get a response from his replacement (CM), so collectively decided to move on. @gkatsev, @eric-carlson, and Ken support the idea, and making the keys case-insensitive will allow it to work for the YouTube use case, so I don't think we need to for an affirmative response from CM.

index.bs
```diff
@@ -4130,6 +4283,30 @@ follows:</p>
...
+
+<ol algorithm="WebVTT attributes block parsing">
+ <li>Let |input| be the list of key/value pairs from a <a>WebVTT attributes block</a>.</li>
+ <li>
+  How the attribute is processed depends on its key name, as follows:
+  <dl>
+
+   <dt>If the key name is "<code>kind</code>"</dt>
```
cookiecrook, line 4294:
```suggestion
   <dt>If the key name is "<code>kind</code>" (case-insensitive)</dt>
```

index.bs
```diff
@@ -4130,6 +4283,30 @@ follows:</p>
...
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>lang</code>"</dt>
```
cookiecrook, line 4297:
```suggestion
   <dt>If the key name is "<code>lang</code>" (case-insensitive)</dt>
```

index.bs
```diff
@@ -4130,6 +4283,30 @@ follows:</p>
...
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang">the srclang attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>label</code>"</dt>
```
cookiecrook, line 4300:
```suggestion
   <dt>If the key name is "<code>label</code>" (case-insensitive)</dt>
```

index.bs
cookiecrook: @gkatsev is this the ID you intended?
```suggestion
<h3 id="rules-for-parsing-attr-key-values-algorithm">WebVTT Attributes key/value Parsing Rules</h3>
<p>The <dfn>WebVTT attributes key/value parsing rules</dfn> consist of the following algorithm.</p>
```

index.bs
cookiecrook: Added "(case-insensitive)" to allow YouTube's use case without modification.

cookiecrook (commented):
Incorporating the YouTube allowance and seeking clarification on a comment from @gkatsev

index.bs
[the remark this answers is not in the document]
cookiecrook: @gkatsev Thought about this some more, and it may help to keep them? Most people don't know the difference between captions and subtitles, so the two examples may provide that clarity.

index.bs
```diff
@@ -1650,6 +1764,45 @@ SIGN).</p>
...
+        <ul>
+         <li>Any <a href="https://infra.spec.whatwg.org/#ascii-alphanumeric">ASCII Alphanumeric</a> character</li>
+         <li>U+005F LOW LINE ("_" underscore)</li>
+        </ul>
+       </li>
+      </ol>
+     </li>
+     <li>A single U+003A COLON character ("<code>:</code>").</li>
+     <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
+     <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
```
aphillips, line 1794: **_Any_** characters except SPACE/TAB/CR/LF and `-->`? That leaves a lot of characters that might potentially be used in an unsavory manner.

index.bs
```diff
@@ -4130,6 +4283,31 @@ follows:</p>
...
+<h3 id="rules-for-parsing-attr-key-values-algorithm">WebVTT Attributes key/value Parsing Rules</h3>
+<p>The <dfn>WebVTT attributes key/value parsing rules</dfn> consist of the following algorithm.</p>
+
+<ol algorithm="WebVTT attributes block parsing">
+ <li>Let |input| be the list of key/value pairs from a <a>WebVTT attributes block</a>.</li>
+ <li>
+  How the attribute is processed depends on its key name, as follows:
+  <dl>
+
+   <dt>If the key name is "<code>kind</code>" (case-insensitive)</dt>
```
aphillips, line 4295: You should define "(case-insensitive)" because it turns out that there are potential variations. Since the thing you're comparing to is an ASCII string, use Infra's "ASCII case insensitive" for best results.

aphillips (changes requested):

index.bs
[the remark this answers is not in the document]
cookiecrook: I'm moving the additional examples to a new intro informative section on Attributes in the Introduction, which looks much better than tacking on additional "other" examples.

index.bs
```diff
@@ -658,6 +714,32 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+
+
+<div class="example">
```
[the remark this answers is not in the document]
cookiecrook: Moving to a new section specifically on the attributes block.

index.bs
cookiecrook: what do you recommend?

aphillips: > what do you recommend?

Let me see if I understand the requirements. 
* The attributes are effectively a "properties file", with constrained keys but allowing a wide range of data
   * WebVTT won't perform matching on the attribute value, but the attribute value consumer might. 
* The value must parse as a single line (that's why CR/LF are escaped)
* I misread before: you do allow space and tab, including at the start (a leading space after the `:` is consumed by the parser, but not others)
   * The intention is thus to allow nearly any text value, including multiline, so long as the CR/LFs are escaped
   
This would allow an attributes to be things like:

```
edited_by: Joe
edited_by: चंदा
```

```
complicated:         فَاطِمَة&#x0A;<--
```

If I understand the requirements right...

I think maybe not much change after all. 

The main problem you might want to guard against is directional spoofing. See [UTR#55: Source Code Handling](https://www.unicode.org/reports/tr55/) for details and examples. Bidirectional formatting characters U+202B, U+202E, U+200F, and U+2067 can all be used to make the sequence `<--` look like:

```

‫<--‬
‮<--
⁧<--⁩

```

So perhaps require the bidi formatting characters in ranges U+202B..202E and U+2066-2069, plus U+200E, U+200F, and U+061C be escaped. Failing that, provide a health warning. WebVTT parsers won't be fooled, but such exploits can fool human users.

index.bs
```diff
@@ -1650,6 +1771,45 @@ SIGN).</p>
...
+      <ul>
+       <li>Any <a href="https://infra.spec.whatwg.org/#ascii-alphanumeric">ASCII Alphanumeric</a> character</li>
+       <li>U+005F LOW LINE ("_" underscore)</li>
+      </ul>
+     </li>
+    </ol>
+   </li>
+   <li>A single U+003A COLON character ("<code>:</code>").</li>
+   <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
+   <li>A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than unescaped U+000A LINE FEED (LF) characters and unescaped U+000D CARRIAGE RETURN (CR) characters, except that the entire resulting string must not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
```
cookiecrook, line 1805: @aphillips wrote: (but this comment dropped out due to a later commit)

> The main problem you might want to guard against is directional spoofing. See UTR#55: Source Code Handling for details and examples. Bidirectional formatting characters U+202B, U+202E, U+200F, and U+2067 can all be used to make the sequence <-- look like:
> 
> ‫<--‬
> ‮<--
> ⁧<--⁩
> 
> So perhaps require the bidi formatting characters in ranges U+202B..202E and U+2066-2069, plus U+200E, U+200F, and U+061C be escaped. Failing that, provide a health warning. WebVTT parsers won't be fooled, but such exploits can fool human users.

cookiecrook, line 1805: > Let me see if I understand the requirements.
> 
> The attributes are effectively a "properties file", with constrained keys but allowing a wide range of data
> WebVTT won't perform matching on the attribute value, but the attribute value consumer might.
> The value must parse as a single line (that's why CR/LF are escaped)
> I misread before: you do allow space and tab, including at the start (a leading space after the : is consumed by the parser, but not others)
> The intention is thus to allow nearly any text value, including multiline, so long as the CR/LFs are escaped

Yes. Exactly.

> This would allow an attributes to be things like:
> 
> edited_by: Joe
> edited_by: चंदा
>  complicated:         فَاطِمَة&#x0A;<--
> If I understand the requirements right...
> 
> I think maybe not much change after all.

Great!

> The main problem you might want to guard against is directional spoofing. See UTR#55: Source Code Handling for details and examples. Bidirectional formatting characters U+202B, U+202E, U+200F, and U+2067 can all be used to make the sequence <-- look like: -->
> 
> So perhaps require the bidi formatting characters in ranges U+202B..202E and U+2066-2069, plus U+200E, U+200F, and U+061C be escaped. Failing that, provide a health warning. WebVTT parsers won't be fooled, but such exploits can fool human users.

Are you suggesting I add those to the exclusion list, or are there legitmate scenarios where those characters should remain allowed in a value? Thanks!

index.bs
cookiecrook: Convo continued at https://github.com/w3c/webvtt/pull/523#discussion_r2051237643

index.bs
aphillips: There are lots of places where these characters are fundamentally needed at render time. However, permitting them unescaped in VTT source can result in spoofing attacks. I'd maybe classify them with CR/LF. Or, if you're not worried about Trojan Source, perhaps do nothing (although I think you might include a health warning in that case).

index.bs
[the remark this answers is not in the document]
nigelmegitt: >Most people don't know the difference between captions and subtitles

[aside]

A significant number of people don't subscribe to the distinction made in the web context between them, and actively use different definitions for the terms. My position on this is that the web usage attempts to bring clarity, but in doing so only brings more problems. IMO it would be better to be explicit about the type of subtitles/captions rather than keep flogging this!

index.bs
cookiecrook: Modifying to the following:
```html
   <li>
    A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than the following:
    <ul>
     <li>unescaped U+000A LINE FEED (LF) characters,</li>
     <li>unescaped U+000D CARRIAGE RETURN (CR) characters,</li>
     <li>bi-directional formatting characters U+202B, U+202C, U+202D, U+202E, U+2066, U++2067, U++2068, U+2069, U+200E, U+200F, U+061C, or</li>
     <li>the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
    </ul>
   </li>
```
I'm hopeful that resolves your concern, @aphillips. Thanks!

index.bs
```diff
@@ -563,6 +651,8 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+
+
```
cookiecrook, line 655:
```suggestion
```

index.bs
```diff
@@ -362,12 +362,100 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+
 <h3 id=introduction-other-features>Other caption and subtitling features</h3>
 
 <p><i>This section is non-normative.</i></p>
 
 <p>WebVTT also supports some less-often used features.</p>
 
+
```
cookiecrook, line 458:
```suggestion
```

cookiecrook (commented):
erroneous whitespace trims

silviapfeiffer:
FWIW I think this is long overdue and the case-insensitive approach will match with the YouTube implementation, so I am all for merging it.

cookiecrook:
@nigelmegitt wrote:

> Looking at it now, does it fix/align with/not overlap with/not clash with https://github.com/w3c/webvtt/issues/485 or is there an additional PR needed? 

No additional PR needed, as far as I can tell. I have suggested https://github.com/w3c/webvtt/issues/485 be closed as a duplicate of the follow-on issue #512. 

> I'm no longer sure of the scope of what this is addressing, since https://github.com/w3c/webvtt/issues/485 is not mentioned in the pull request description but has clearly been at least taken into account in edits made since I last looked.

I was not previously aware of @yashrajbharti's issue, so "taken into account" was merely serendipitous, but I do believe it accounts for all the prerequisites needed by #512 and https://github.com/w3c/webvtt/issues/485

If there are no further review comments on this one, please merge and I'll start the follow-on work for:
- #512

gkatsev (approved):

index.bs
```diff
@@ -1650,6 +1767,53 @@ SIGN).</p>
...
+    </ol>
+   </li>
+   <li>A single U+003A COLON character ("<code>:</code>").</li>
+   <li>Zero or one U+0020 SPACE or U+0009 CHARACTER TABULATION (tab) characters.</li>
+   <li>
+    A <dfn>WebVTT attribute value</dfn> consisting of any sequence of zero or more characters other than the following:
+    <ul>
+     <li>unescaped LINE FEED (LF) characters (U+000A),</li>
+     <li>unescaped CARRIAGE RETURN (CR) characters (U+000D),</li>
+     <li>unescaped bi-directional formatting characters (U+202B, U+202C, U+202D, U+202E, U+2066, U++2067, U++2068, U+2069, U+200E, U+200F, U+061C), or</li>
```
nigelmegitt, line 1806: No escaping mechanism appears to be defined.

index.bs
```diff
@@ -4130,6 +4294,34 @@ follows:</p>
...
+  <dl>
+
+   <dt>If the key name is "<code>kind</code>" (<a href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII case-insensitive</a>)</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>lang</code>" (<a href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII case-insensitive</a>)</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-srclang">the srclang attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>label</code>" (<a href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII case-insensitive</a>)</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-label">the label attribute</a> of a track element according to the HTML Standard.</dd>
```
nigelmegitt, line 4313: When this says "Process the value as a ... according to the HTML standard", what is meant? Is it only processing it as a syntactic entity, or is there an expectation that the contents of the file will be handled differently.

What is supposed to happen if a `<track>` element has one or more of these keys as attributes, and their values differ from the values in a referenced WebVTT file? Is that an error? Which takes precedence?

nigelmegitt (commented):
I'm not sure this is ready to merge yet; I've left some questions.

zcorpan (changes requested):
This does not define parsing of ATTRIBUTES blocks in https://w3c.github.io/webvtt/#file-parsing , so currently they are dropped during parsing.

Why not use the "header" parser logic for attributes? i.e.

```
WEBVTT
kind: subtitles
lang: es-mx
label: Español

...
```

index.bs
gkatsev: > WebVTT should make the data available and HTML should say what to do with it in its track processing model.


I believe that was the intent.

> What is supposed to happen if a `<track>` element has one or more of these keys as attributes, and their values differ from the values in a referenced WebVTT file? Is that an error? Which takes precedence?

I would expect the `<track>` element attribute to win in this case. But it's probably up to how it'll get defined in an HTML update.


"Process the value as" is probably not clear enough as to what we expect to happen here. My expectations are this:
- link `label`, `lang`, and `kind` to the definitions in HTML for `label`, `srclang`, and `kind`
- have these values available in such a way that HTML can use them directly without extra processing

zcorpan: > If we're expecting changes to HTML to inspect the referenced WebVTT files and use their values to populate the `kind`, `label` and `srclang` attributes

I would not expect that to happen, but the format-provided values would populate the internal concepts and be readable from the `TextTrack` object.

zcorpan:
OK but per https://github.com/w3c/webvtt/issues/485#issuecomment-3084770525 it seems YouTube already uses the header for these attributes?

HLS timestamp map uses different keys, so it doesn't conflict, as far as I can tell.

index.bs
```diff
@@ -4130,6 +4294,34 @@ follows:</p>
...
+<ol algorithm="WebVTT attributes block parsing">
+ <li>Let |input| be the list of key/value pairs from a <a>WebVTT attributes block</a>.</li>
+ <li>
+  How the attribute is processed depends on its key name, as follows:
+  <dl>
+
+   <dt>If the key name is "<code>kind</code>" (<a href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII case-insensitive</a>)</dt>
+   <dd>Process the value as <a href="https://html.spec.whatwg.org/multipage/media.html#attr-track-kind">the kind attribute</a> of a track element according to the HTML Standard.</dd>
+
+   <dt>If the key name is "<code>lang</code>" (<a href="https://infra.spec.whatwg.org/#ascii-case-insensitive">ASCII case-insensitive</a>)</dt>
```
zcorpan, line 4309: Why is this `lang` and not `language`, to match https://github.com/w3c/webvtt/issues/485#issuecomment-2362201249 ?

index.bs
zcorpan: I've filed https://github.com/whatwg/html/issues/11665

zcorpan (changes requested):
Marking as request changes to address my comments.

gkatsev:
looks like webm also has file-wide metadata https://wiki.webmproject.org/webm-metadata/temporal-metadata/webvtt-in-webm
it seems it uses Language and Kind there, though, on another page it uses VTTKind: https://wiki.webmproject.org/webm-metadata/temporal-metadata/webvtt-metadata
it also looks like the values they use are capitalized.

Looking at the HLS spec again, I guess the main difference is that this block is currently define as `key: value` but in HLS with the X-TIMESTAMP-MAP, it's `X-TIMESTAMP-MAP=value`. Do we want to account for both `:` and `=` as the delimiter here?

Also, if we do move it back to the pre-existing headers location, which would fully resolve #485, the HLS spec would likely need to get updated to account for potential other items in the headers.

---------

Also, looking at the webm pages, there's a link to https://lists.w3.org/Archives/Public/public-texttracks/2012Apr/0011.html, which has in interesting suggestion of a `getHeaders()` API. That sounds like a great idea, but I don't think it needs to get added as part of this PR specifically.

kathleenmotley49-maker (approved):

index.bs
```diff
@@ -1436,6 +1550,70 @@ together, which is particularly important when scrolling up.</p>
...
+  "<code>custom</code>" or a string beginning with "<code>custom-</code>". All other values
+  are reserved for future standardization. Defaults to the empty string.</p>
+  <p class="note">The <code>type</code> key disambiguates the track kind subtype to resolve
+  naming conflicts for the other common key names often used by different types of metadata.</p>
+  <p class="note">Authors including
+  <a lt="WebVTT attributes object custom pairs">custom pairs</a> should provide a non-empty
+  <code>type</code> value to identify the application or schema those pairs belong to.
+  A <a>WebVTT attributes block</a> with non-empty
+  <a lt="WebVTT attributes object custom pairs">custom pairs</a> and an empty
+  <code>type</code> is valid but parsers may generate a warning.</p>
```
cookiecrook: Same as above for this warning... I'm also open to cutting the warning if reviewers feel it's not necessary.

cookiecrook (commented):
Notes for other reviewers, and a couple changes I noticed after the latest PR push.

index.bs
```diff
@@ -1436,6 +1550,70 @@ together, which is particularly important when scrolling up.</p>
...
+ </dd>
+
+ <dt><dfn lt="WebVTT attributes object language">A language</dfn></dt>
```
cookiecrook: Open question here re: consistency of `language` vs `lang` vs `srclang`.
https://github.com/w3c/webvtt/pull/523#issuecomment-4034078953

index.bs
cookiecrook: To clarify, are you suggesting:

1. Leaving the author SHOULD (1578–1580) but dropping the parser/validator MAY (1581–1583), or 
2. Dropping both?

zcorpan:
I still think this should be in the header instead of a new block (see https://github.com/w3c/webvtt/pull/523#pullrequestreview-3224169912 ).

zcorpan:
> **Open Question:** I changed the `lang` key to `language` based on review feedback here and precedent for YouTube's usage of `Language: en`... However, this now conflicts with WebVTT's `<lang en>` element, as well as HTML's `lang` and `srclang` attributes. I don't personally care which the WebVTT spec uses for the new `ATTRIBUTES` block key, so I'd like the Timed Text Working Group to make the decision.

`lang` is certainly more consistent, and looks nicer with `kind` and `type` also being 4 characters long. I suppose Youtube should be able to change to conform.

cookiecrook:
### Open Questions for TTWG Meeting Discussion (Date TBD)

1. **Should this use the new `ATTRIBUTES` block as proposed vs. putting the key/value pairs in the existing header.**
	- @zcorpan thinks it should be in the header
	- ~~I’m strongly for~~ **_I previously was for_** the ATTRIBUTES block for transitional ease wrt existing, non-conforming use of header parsing. (Note: @zcorpan objected to my use of “webcompat” above since the existing header usage, although widespread, is already non-conforming and therefore not technically a webcompat concern.)
	- Other TTWG members (including but not limited to @gkatsev) have expressed interest in the ATTRIBUTES, but I don’t know how strongly this belief is held or by whom.

2. **Key name parsing, including character: define here, reference HTML, or use another name definition?**
	- ASCII+ (as in PR), XML Name, or HTML attribute name?
	- Some characters need exclusion.
		- Bidi and reverse chars excluded.
		- Are we concerned with look-alike character like those comparable in ASCII and Cyrillic? 
		- What about zero-width joiners used in emoji variant combinations?

3. **Should the reserved key name be lang or language?**
	- Previously `lang`, changed to `language` at @nigelmegitt’s review request (Note: my PR misses that change in at least one place)
	- @zcorpan prefers lang... I have a mild pref for lang, too.
	- No one has voted for `srclang` the other existing precedent.
	- Not sure about others.

4. **Should `kind` be required, or should `kind: metadata` be required in order to use (metadata) `type`?**
	- Sounds like no, but I’d like to discuss anyway.
	- See follow-on goal for metadata type in https://github.com/w3c/webvtt/pull/523#discussion_r2921729393 
	- See also 5 below.

5. **Should there be any parser warning, or should it be left to a validator, rather than the parser?**
	- Sounds like there is mild agreement that there is no need for a parser warning.
	- Depending on outcome of 4 above, this may be moot.

6. **Remove the `subtitles` example since it is so similar to the `captions` example.** 
	- The TTWG feels the difference between subtitle (for those who can hear but need translation of the primary language) vs captions (which includes sensory needs like Deaf and Hard-of-Hearing groups, as well as those with translation needs) is not relevant to call out in VTT spec examples.
	- I have a mild pref to keep both, but will concede to the TTWG consensus after discussion.

cookiecrook:
@nigelmegitt @gkatsev @zcorpan Please help if I missed anything in summary of open questions above. Thanks.

I know that @eric-carlson would like to attend said meeting, and is out this week. My next availability at the normal TTWG meeting time (8am Pacific) is April 9th or April 23rd.

gkatsev:
My next availability is also April 9th.

I still need to review the new discussion but should note that I'm leaning towards using the headers space assuming that we allow non-conforming usage like mentioned here https://github.com/w3c/webvtt/pull/523#issuecomment-3300189061

nigelmegitt:
Agenda for TTWG 2026-04-09 is w3c/ttwg#331, which I'm unable to attend, but have added this as a possible agenda item. 23rd April would work for me, alternatively.

zcorpan:
> * (Note: @zcorpan objected to my use of “webcompat” above since the existing header usage, although widespread, is already non-conforming and therefore not technically a webcompat concern.)

To clarify, I objected to the claim that existing WebVTT content would go from being conforming to being non-conforming, as it's currently invalid WebVTT to use non-blank lines immediately after the `WEBVTT` file signature.

There could be a webcompat concern, but it's separate from whether existing content is conforming or non-conforming. I asked for examples where the new processing of headers would break existing content. The fact that existing content is using headers doesn't mean that that content will break if browsers start to process some known headers with some known values. I think there's only a *potential* issue if the standardized key and value are exactly the same as what already exists in the wild, but the semantics are different.

zcorpan:
> ASCII+ (as in PR), XML Name, or HTML attribute name?

If we don't need to allow characters outside of ASCII, an option is to align with HTML PI target: https://github.com/whatwg/html/pull/12118

Right now it's `[a-zA-Z][a-zA-Z0-9-]*` but I added a review comment asking if `_` should also be allowed.

nigelmegitt:
> If we don't need to allow characters outside of ASCII

Since the metadata keys may originate or relate to schemas external to WebVTT and to HTML, my view is that only allowing ASCII characters is likely to be overly restrictive. It's also unnecessary to be so limiting.

cookiecrook:
Since Nigel can't make it on April 9th, can everyone make it on April 23rd?

If not, let's go ahead with April 9th.

gkatsev:
Yeah, I think the 23rd works a bit better for me as well. We're going to cancel the April 9th meeting and have the agenda item be on the 23rd.

index.bs
```diff
@@ -362,6 +362,93 @@ CSS comment (e.g. <code>/**/</code>).</p>
...
+translate spoken dialog or signage, but not audible sound 
+effects like "dogs barking."
+
+1
+00:00:10.123 --> 00:00:15.432
+¡Hola! ¿Qué tál?
+ </pre>
+
+</div>
+
```
cookiecrook: Deleting as per review comment in: https://github.com/w3c/webvtt/pull/523#discussion_r2925086204
```suggestion
```

index.bs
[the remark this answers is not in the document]
cookiecrook: Resolved in https://github.com/w3c/webvtt/pull/523/changes/158e2125d27b2dd64369cb81ee38857d9cdad649

gkatsev:
Some comments on this ahead of the meeting.

> ### Open Questions for TTWG Meeting Discussion (Date TBD)
> 
>     1. **Should this use the new `ATTRIBUTES` block as proposed vs. putting the key/value pairs in the existing header.**
>        
>        * @zcorpan thinks it should be in the header
>        * I’m strongly for the ATTRIBUTES block for transitional ease wrt existing, non-conforming use of header parsing. (Note: @zcorpan objected to my use of “webcompat” above since the existing header usage, although widespread, is already non-conforming and therefore not technically a webcompat concern.)
>        * Other TTWG members (including but not limited to @gkatsev) have expressed interest in the ATTRIBUTES, but I don’t know how strongly this belief is held or by whom.

Originally, I thought a new block would be better to minimize issues, but if we can support both the syntax we want while making existing non-conforming usage conforming (albeit a non-standard usage) , I think that would be best. I'm not sure if there's precedent for this type of thing in the W3C, but for JavaScript, there's some precedent when they standardized the previously non-standard `__proto__` property.

>     2. **Key name parsing, including character: define here, reference HTML, or use another name definition?**
>        
>        * ASCII+ (as in PR), XML Name, or HTML attribute name?
>        * Some characters need exclusion.
>          
>          * Bidi and reverse chars excluded.
>          * Are we concerned with look-alike character like those comparable in ASCII and Cyrillic?
>          * What about zero-width joiners used in emoji variant combinations?

Is there a need to limit it to ASCII+ specifically? For example, the cue identifier says it's anything except --> and it also specifies that it must be unique, so, theoretically, unicode normalization and what not should already be happening.

```
A WebVTT cue identifier is any sequence of one or more characters not containing the substring "-->" (U+002D HYPHEN-MINUS, U+002D HYPHEN-MINUS, U+003E GREATER-THAN SIGN), nor containing any U+000A LINE FEED (LF) characters or U+000D CARRIAGE RETURN (CR) characters.

A [WebVTT cue identifier](https://w3c.github.io/webvtt/#webvtt-cue-identifier) must be unique amongst all the [WebVTT cue identifiers](https://w3c.github.io/webvtt/#webvtt-cue-identifier) of all [WebVTT cues](https://w3c.github.io/webvtt/#webvtt-cue) of a [WebVTT file](https://w3c.github.io/webvtt/#webvtt-file).
```
https://github.com/w3c/webvtt/blob/99666098eef0c025c2ab5f3d2b12bf94521aae21/index.bs#L1547-L1553

Also, the VTT Region identifier is similar and disallows spaces.

Also, should this be case-insensitive?

>     3. **Should the reserved key name be lang or language?**
>        
>        * Previously `lang`, changed to `language` at @nigelmegitt’s review request (Note: my PR misses that change in at least one place)
>        * @zcorpan prefers lang... I have a mild pref for lang, too.
>        * No one has voted for `srclang` the other existing precedent.
>        * Not sure about others.

I do prefer lang as well. Is there a case to support both and have it normalized internally? Being case-insensitive and supporting both would allow supporting the YT attr without work on their side.
 
>     4. **Should `kind` be required, or should `kind: metadata` be required in order to use (metadata) `type`?**
>        
>        * Sounds like no, but I’d like to discuss anyway.
>        * See follow-on goal for metadata type in [New ATTRIBUTES block #523 (comment)](https://github.com/w3c/webvtt/pull/523#discussion_r2921729393)
>        * See also 5 below.

Having a type could imply a metadata type, though, probably not worth adding. If we'd want to extend type for other kinds, it might make it a bit harder. 

>     5. **Should there be any parser warning, or should it be left to a validator, rather than the parser?**
>        
>        * Sounds like there is mild agreement that there is no need for a parser warning.
>        * Depending on outcome of 4 above, this may be moot.

Yeah, I think I would prefer to err on the permissive side to have less requirements and less parser errors, if it doesn't cause any major issues.

>     6. **Remove the `subtitles` example since it is so similar to the `captions` example.**
>        
>        * The TTWG feels the difference between subtitle (for those who can hear but need translation of the primary language) vs captions (which includes sensory needs like Deaf and Hard-of-Hearing groups, as well as those with translation needs) is not relevant to call out in VTT spec examples.
>        * I have a mild pref to keep both, but will concede to the TTWG consensus after discussion.

could alternate kinds across different examples instead, maybe? Like, update some of the existing examples with the new attrs.

cookiecrook:
### Consensus from today's meeting as I understand it.

1. key/value pairs should immediately follow the `WEBVTT` file header, not a new `ATTRIBUTES` block.
2. Allowed key name characters will match the value: any unicode char minus exclusions such as bidi or reverse chars, `-->`, etc.
3. language key should be `lang` to match HTML `<track lang>`
4. Neither `kind` nor (metadata) `type` should be required at this stage.
5. No parser warnings needed.
6. I already removed the subtitles example due to its similarity to the captions example.

#### Consensus of other items discussed:

7. Make key names case-sensitive; existing implementations can update.
8. In addition to colon (`:`) for key/value delineation, allow equals (`=`) for webcompat related to HLS VTT's pervasive use of `X-TIMESTAMP-MAP=`, but mark equals as not recommended for other use.
9. In the case of a mismatch between a standardized attribute from HTML (E.g. `<track lang>` vs the file's `lang` value), the value from HTML will prevail as source of truth. (**Follow-up:** We did not discuss inherited values vs explicit values though. Would an inherited `:lang(en)` from the document count as an override, or only a non-empty `lang` value on `<track>`?)
10. Any invalid key/value lines are ignored, but an invalid pair will not invalidate the whole block.
11. Custom keys (those not defined by the VTT spec or a registry `kind->type` extension spec) MUST include a hypen (`-`) in the name such as `x-foo: bar` (or `data-bop: baz` to match HTML)... (**Follow-up:** We did not discuss if a leading hyphen `-foo` would be okay.)

Please fill in if I've missed anything. Thanks everyone for your time today.

cookiecrook:
I also plan to close this PR and create a new one with the rewrite because the thread is getting unmanageably long.

gkatsev:
Thanks for the summary! Seems accurate to what was discussed.

For item 9, there's probably nothing needed on the WebVTT spec side as it'll be part of HTML or related specs.

For item 11 follow-up, might be fine to not worry if it starts with a `-`. The only reason I can think of requiring text at the beginning is if we wanted `getAttributes()` to return an object and handle it similarly to CSS where we set the keys of that object to be camelCased versions of the attribute name in the file. But, that's probably unneeded as we can return the name as a string.

gkatsev:
We discussed item 11 briefly at today's TTWG meeting. It'll be good to see how CSS handle's the camelCasing of properties that start with `-`. Also, that it might be better to disallow starting with `-` at the moment, as it's always easier to relax requirements later on rather than tighten requirements if we find issues.

cookiecrook:
This pull request:
- #523 

Has been superseded by:
- #548

Which is currently a **_draft_** PR. I will mark as non-draft PR when it's ready for wider review.

cookiecrook:
Closing this as superseded by PR #548 which is ready for review:
- https://github.com/w3c/webvtt/pull/548

PR #528, opened 2025-02-12
Add support for ::cue-container

### LINKED ISSUE #516: WebVTT should expose an inline-block backdrop element for cues
Safari and WebKit on macOS and iOS have a long-standing default treatment of text track cues that incorporates an `inline-block` wrapper around the cues, so as to provide the same rendering of cues as is available to native applications using system media frameworks. The styles for this `inline-block` wrapper are configurable through system defaults, allowing the user to choose between an `inline-block` style highlight, and a `inline` highlight. (See attached images for examples.)

Because this block is not exposed through WebVTT, it is very difficult (if not impossible) for inline- or page-provided styles to override this rendering.

inline-block
inline

jernoble:
Fixes #516

In order to align the capabilities of WebVTT with CEA standards like 708, add the concept of a "WebVTT block background box", which is defined as a block-level wrapper around the WebVTT cue's contents.

To allow authors to specify styles that apply to this block background object, add a new selector `::cue-container` that matches it. UAs which provide FCC required styling information for "window background & opacity" can do so via this new selector in UA-provided style sheets.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4786,7 +4786,11 @@ constraints: [[!CSS22]]</p>
  sheets after their boxes are generated, as described below.)</li>
 
  <li>The children of the |nodes| must be wrapped in an anonymous box whose 'display' property has
- the value ''display/inline''. This is the <dfn>WebVTT cue background box</dfn>.</li>
+ the value ''display/inline''. This is the <dfn>WebVTT cue inline background box</dfn>.</li>
+
+ <li>The <a>WebVTT cue inline background box</a> must be wrapped in an anonymous box whose 'display'
+ property has the value ''display/block''. This is the <dfn>WebVTT cue block background
+ box</dfn>.</li>
 
  <li>Runs of children of <a lt="WebVTT Ruby Object">WebVTT Ruby Objects</a> that are not <a
  lt="WebVTT Ruby Text Object">WebVTT Ruby Text Objects</a> must be wrapped in anonymous boxes whose
@@ -4873,8 +4877,8 @@ corresponding <a lt="text track cue">cue</a>'s <a>WebVTT cue text alignment</a>:
 <p>The 'color' property on the (root) <a>list of WebVTT Node Objects</a> must be set to
 ''rgba(255,255,255,1)''. [[!CSS3-COLOR]]</p>
 
-<p>The 'background' shorthand property on the <a>WebVTT cue background box</a> and on <a>WebVTT Ruby
-Text Objects</a> must be set to ''rgba(0,0,0,0.8)''. [[!CSS3-COLOR]]</p>
+<p>The 'background' shorthand property on the <a>WebVTT cue inline background box</a> and on
+<a>WebVTT Ruby Text Objects</a> must be set to ''rgba(0,0,0,0.8)''. [[!CSS3-COLOR]]</p>
 
 <p>The 'white-space' property on the (root) <a>list of WebVTT Node Objects</a> must be set to
 ''white-space/pre-line''. [[!CSS22]]</p>
@@ -5302,10 +5306,10 @@ pseudo-elements defined below won't have any effect according to this specificat
 
 <h4 id=the-cue-pseudo-element>The ''::cue'' pseudo-element</h4>
 
-<p>The <dfn selector>::cue</dfn> pseudo-element (with no argument) matches any <a>list of WebVTT Node
-Objects</a> constructed for the <i>matched element</i>, with the exception that the properties
-corresponding to the 'background' shorthand must be applied to the <a>WebVTT cue background box</a>
-rather than the <a>list of WebVTT Node Objects</a>.</p>
+<p>The <dfn selector>::cue</dfn> pseudo-element (with no argument) matches any <a>list of WebVTT
+Node Objects</a> constructed for the <i>matched element</i>, with the exception that the properties
+corresponding to the 'background' shorthand must be applied to the <a>WebVTT cue inline background
+box</a> rather than the <a>list of WebVTT Node Objects</a>.</p>
 
 <p>The following properties apply to the ''::cue'' pseudo-element with no argument; other properties
 set on the pseudo-element must be ignored:</p>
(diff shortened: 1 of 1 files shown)


jernoble:
Pinging @gkatsev for a review of this PR.

gkatsev:
This change seems reasonable to me. Though, we'd want to resolve naming issues between this one and https://github.com/w3c/webvtt/issues/531. Maybe this should be `::cue-backdrop` and #531 the `::cue-container`?

nigelmegitt:
I think `::cue-backdrop` is not a great name since it creates an ambiguity: does it apply at the block level or the inline level. Maybe `::cue-block` would work better, allowing for a potential `::cue-inline` if ever needed later?

jernoble:
Our accessibility and CSS experts internally suggest "block background" and "inline background" as naming conventions. So this would imply `::cue-block-background` and (in the future) `::cue-inline-background`.

nigelmegitt:
That makes sense @jernoble , except... isn't the `::cue-thing` part of a selector, to target a particular stylable thing, and if you want to style the background, you'd apply the CSS `background-color` property? 

If we ever wanted to allow other styling properties to apply, it might look rather odd to have `background` be part of the selector; putting something that looks like a property name into a selector is probably not ideal, even if it might be unavoidable sometimes.

jernoble:
> If we ever wanted to allow other styling properties to apply, it might look rather odd to have background be part of the selector...

Well yes, which is why I originally proposed "container" or "backdrop", which are both descriptive of the thing itself, and not the property. I don't actually think "block" or "inline" alone are appropriate either for the same reason: those are both CSS `display:` values (in addition to being CSS concepts). Besides, the "block" object is actually `display:inline-block`.

nigelmegitt:
Aargh, naming is hard! Yes, I agree with those concerns too, though I feel that overlap with a permitted property _value_ is less bad than overlap with a property _name_.

jernoble:
If you wanted to disambiguate the terms from their CSS property values, one option would be to use [`inline box`](https://drafts.csswg.org/css-display/#inline-box) and [`block container`](https://drafts.csswg.org/css-display/#block-container), both of which are terms defined by [CSS](https://www.w3.org/TR/css-display-3/), but are not themselves CSS property values.

jernoble:
Or, we could roll all the way back to `::cue-container` (since "block container" has a defined meaning in CSS), and use `::cue-root` (since "root element" also has a defined meaning in CSS) to represent the cue display area, and the use of those two terms shouldn't be ambiguous (in CSS terms).

gkatsev:
I guess `::cue-block-box` could be an option, but maybe `::cue-container` with `::cue-root` might be best/easiest. I don't really have other ideas.
Do we know folks from Firefox or Chrome who might have thoughts?

fantasai:
+1 to Nigel's comments wrt incorporating -background in the name. :)

What is the box-tree you're working with here? If I set `color` on this pseudo-element, will it inherit to `::cue` elements inside it? If I size it, will that affect the content inside it?

If no, then `backdrop` is fairly reasonable as a name. But if it does function as a container, then `::cue-block` might work.

jernoble:
> If I set color on this pseudo-element, will it inherit to ::cue elements inside it?

You cannot set `color:` on it, because `color:` isn't an allowed property on this selector. 

> If I size it, will that affect the content inside it?

The only allowed properties would be `background:`, `padding:` and `border:`; setting those would affect the layout of its contents, but not their size.

jernoble:
Per @gkatsev 's comment in [issue 531](https://github.com/w3c/webvtt/issues/531#issuecomment-2725945333), updating this PR to use `:cue-backdrop`.

alastor0325:
I've reviewed this issue and #516. Adding a new pseudo-element for the block cue window, as well as its proposed naming, both sound reasonable to us (Mozilla).

However, we have a concern regarding how this proposal involves applying system accessibility settings for closed captions. It seems possible for websites to infer user-specific system styles from the computed cue styles, which could lead to fingerprinting risks. How should we address this potential privacy issue?

Additionally, if system settings are applied to cue rendering and a website explicitly sets cue colors that happen to match those from the system settings (for example, both text and background being white), it could result in even worse accessibility for users. How should cases like this be handled?

emilio:
Yeah two issues that immediately came to mind. For the example, let's say system a11y settings have effectively `background-color: white` and `color: black`.

Issue number 1 is causing unreadable backdrops on existing content. Let's say I have a dark video, so as an author unaware of these settings, I write:

```css
video::cue {
  color: white;
}
```

This works if the backdrop is transparent, but a user with such settings would get white on white and unreadable text.

The other issue is that if you just apply the OS setting to the container you can trivially use `getComputedStyle(video, "::cue-backdrop").backgroundColor` to extract the OS setting values, right? That would be a fingerprinting concern.

jernoble:
There is no way to extract style information from inside the shadowDOM of the media element, or that privacy risk would already exist, since both chrome and Safari both already apply system accessibility preferences to browser rendered subtitles.

jernoble:
> ...getComputedStyle(video, "::cue-backdrop").backgroundColor to extract the OS setting values, right?

That won't do anything interesting; `::cue-backdrop` isn't something that can be querySelector'd.

jernoble:
> @jernoble that's generally not right?

This may just be a result of WebKit's implementation of Text Track rendering taking place in the media element Shadow DOM. Styles defined inside the closed Shadow DOM are generally not accessible from the outside.

> I mean, we _can_ decide to not expose `::cue-backdrop` to getComputedStyle, but `getComputedStyle(.., "::before")` and so on work and they can't be querySelector'd

It seems like we have an interop problem; Chrome and Safari both do not expose resolved styles of `::cue` pseudos, and both _do_ apply system accessibility settings to rendered cues; Firefox does expose those resolved styles, and does _not_ apply system accessibility settings. I suggest we take that up in a different issue.

jernoble:
@alastor0325 said:

> Additionally, if system settings are applied to cue rendering and a website explicitly sets cue colors that happen to match those from the system settings (for example, both text and background being white), it could result in even worse accessibility for users. How should cases like this be handled?

Lets continue this conversation in a separate issue that's narrowly about applying system accessibility preferences to subtitle rendering.

jernoble:
I filed Issue #536 to track the interop and accessibility discussion.

gkatsev:
@alastor0325 @emilio I guess my question is do you consider these interop issues a blocker for this PR or is this PR otherwise good to go, and we can continue discussing the interop issues in #537?

PR #539, opened 2026-02-15
Use spec-prod for gh-pages publication

himorin:
as title, to make things simple.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,23 +1,26 @@
-name: Build and Deploy
+name: Build, and publish spec to GitHub Pages and /TR/
+
 on:
+  pull_request: {}
   push:
-    branches:
-      - main
+    branches: [main]
+    paths:
+      - 'logo/**'
+      - 'index.bs'
+
 jobs:
-  build-and-deploy:
+  main:
+    name: Build, Validate and Deploy
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout 🛎️
-        uses: actions/checkout@v2.3.1 # If you're using actions/checkout@v2 you must set persist-credentials to false in most cases for the deployment to work correctly.
+      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


himorin:
@nigelmegitt @gkatsev this PR is what I've mentioned (using echidna for streamlined publication to /TR/ as CRD, as well as pushing to github.io) during last call at:
https://www.w3.org/2026/03/12-tt-minutes.html#107e

It seems I need to update some, llike w3cid or deleting more (like what [format.py](https://github.com/w3c/webvtt/blob/429ed009da32fd56a5274a41585a2fa14233ff34/format.py) is for - could not find any relevant use and I may need to read and understand script...), so let me prepare in parallel to 2w CfC period in TTWG.

gkatsev:
Looks like `format.py` is used to format the index.bs file a bit. Like making sure the line-length isn't too long.
Looks like it was written from python 2 but seems to work with python 3 with minimal changes. I'll open a PR for it and we can see if we want to keep it.

himorin:
@gkatsev thank you really much for investigation.
hrm, if it is something to fix and update bikeshed source itself, I somehow wonder if we could have it as check against PR to point potential point to be fixed rather than run as offline/online tooling as automatical fixup. I observe some lint tool running at some other repositories, and I always encounter difficulty to check and track root point of certain changes, e.g. diff over large amount of time may contain normative changes with overrided lint commit.

gkatsev:
yeah, I agree that it's a bit annoying, though, hopefully if things are maintained as clean it helps not have large changes in a single commit. Though, it is possible to [hide commits from blame](https://www.git-tower.com/blog/how-to-exclude-commits-from-git-blame), so, we can always do a fixup in a PR and add that commit to the exclude.

Also, I do agree that if bikeshed has a build-in formatter, it would be nice to use that.

himorin:
closing per #546 
note, `format.py` discussion may be followed up in other issue.

PR #543, opened 2026-03-16
Replace api.csswg.org with spec-generator

tidoust:
The Bikeshed service at api.csswg.org has been deprecated in favor of spec-generator. The former URL redirects to a documentation that explains how to migrate:
https://github.com/w3c/spec-generator/wiki/Migrating-Bikeshed-HTTP-API-usage-to-spec%E2%80%90generator

That redirection unfortunately makes the build somewhat succeeds without error, creating an unhelpful 302 document instead of the actual spec.

This update makes the Makefile use the spec-generator service.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,5 +4,5 @@ all: index.html
 
 # build using spec-generator but not local bikeshed
 index.html: index.bs
-	curl https://www.w3.org/publications/spec-generator/ -F file=@index.bs -F type=bikeshed-spec -F output=error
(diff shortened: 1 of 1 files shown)


deniak (changes requested):

deniak (approved):

himorin:
hi @tidoust , thank you for updating this. TTWG is considering to replace this with echidna based publication and github.io management. I'll work with co-chairs (both WG and CG?) and editors to have (at least) echidna to github.io asap.
https://www.w3.org/2026/03/12-tt-minutes.html#107e

PR #548, opened 2026-05-28
Add WebVTT header metadata to disambiguate metadata cues (#511)

### LINKED ISSUE #511: VTT Metadata Cue format is ambiguous; some metadata may be unintentionally presented to the user in a context outside HTML

cookiecrook:
Replaces the ATTRIBUTES-block approach explored in #523 with the TTWG-consensus design: file-level key/value pairs are written on the lines immediately following the WEBVTT file header line and are terminated by a blank line, rather than living in a separate block.

Key behaviors per the April 23 + May 7 TTWG meeting consensus and follow-up review comments:

- Reserves four header keys: lang, kind, label, type (all optional; lang aligns with HTML <track srclang>; type addresses #511 by identifying metadata schemas, with the taxonomy work continuing in #512).
- Accepts both ":" and "=" as separators; "=" is permitted for HLS-compatible files but is marked non-recommended.
- Matches reserved keys case-sensitively; authors must use lowercase.
- Allows Unicode in keys and values, excluding bidi controls, line breaks, and the "-->" substring.
- Requires non-reserved keys to contain a hyphen and not start with one, reserving the hyphen-free namespace for future standardization.
- Parses leniently: invalid lines are ignored individually and do not invalidate the rest of the header; no parser warnings are required.
- Adds the parser plumbing (a new |header metadata| slot on the parser signature, a "collect WebVTT header metadata" algorithm, and a new return path from "collect a WebVTT block" when the in-header flag is set).
- Updates the WebVTT metadata text prose to recommend declaring kind and type for files delivered outside an HTML context, addressing the cue-format ambiguity from #511.

Closes #511. Defers HTML-integration details (precedence between VTT header metadata and <track> attributes, processing model) to whatwg/html#11665.

Note: GenAI (Claude Opus 4.8) used in the creation of this latest iteration.

***
<a href="https://pr-preview.s3.amazonaws.com/cookiecrook/webvtt/pull/548.html" title="Last updated on Jul 8, 2026, 9:42 PM UTC (39cab0d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webvtt/548/0648025...cookiecrook:39cab0d.html" title="Last updated on Jul 8, 2026, 9:42 PM UTC (39cab0d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1902,9 +1902,7 @@ or end with a <a>WebVTT line terminator</a>.)</p>
 files delivered outside of an HTML context may declare a
 <a lt="WebVTT header metadata kind">kind</a> of "<code>metadata</code>" together with a
 <a lt="WebVTT header metadata type">type</a> identifying the schema in use; see
-[[#header-metadata]]. Consumers that do not recognise the
-<a lt="WebVTT header metadata type">type</a> value must not present the cue payloads as
-caption or subtitle text.</p>
+[[#header-metadata]].</p>
 
 
 <h4 id=caption-text>WebVTT caption or subtitle cue text</h4>
--- a/index.bs
+++ b/index.bs
@@ -1897,7 +1897,7 @@ or end with a <a>WebVTT line terminator</a>.)</p>
 
 <p><a>WebVTT metadata text</a> cues are typically used by scripted applications (e.g. using the
 <code>metadata</code> <a>text track kind</a> in a HTML <a>text track</a>). Authors of metadata
-files delivered outside of an HTML context should declare a
+files delivered outside of an HTML context may declare a
 <a lt="WebVTT header metadata kind">kind</a> of "<code>metadata</code>" together with a
 <a lt="WebVTT header metadata type">type</a> identifying the schema in use; see
 [[#header-metadata]]. Consumers that do not recognise the
--- a/index.bs
+++ b/index.bs
@@ -1666,21 +1666,21 @@ all of the following properties:</p>
  ISOLATE, U+2068 FIRST STRONG ISOLATE, or U+2069 POP DIRECTIONAL ISOLATE.</li>
  <li>It does not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D
  HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
- <li>It must <em>either</em> match a reserved key, <em>or</em> must match the additional 
+ <li>It must <em>either</em> match a reserved key, <em>or</em> must match the additional
   hyphenated formatting requirements of a custom key:
   <ul>
-   <li>reserved key: Matches one of the keys reserved by this specification for <a>WebVTT header 
+   <li>reserved key: Matches one of the keys reserved by this specification for <a>WebVTT header
    metadata</a>; see [[#header-metadata]] for the current list of reserved keys.</li>
-   <li>custom key: contains at least one U+002D HYPHEN-MINUS character, and its first character 
+   <li>custom key: contains at least one U+002D HYPHEN-MINUS character, and its first character
    is not a U+002D HYPHEN-MINUS character.</li>
   </ul>
  </li>
 </ul>
 
-<p class="note">The hyphenated requirement only applies to custom keys that are not reserved by 
+<p class="note">The hyphenated requirement only applies to custom keys that are not reserved by
 this specification, such as "data-foo". It allows application-specific usage while reserving
-the hyphen-free namespace for future standardization (for example, by the taxonomies anticipated 
-in <a href="https://github.com/w3c/webvtt/issues/512">WebVTT issue #512</a>). Reserved keys 
+the hyphen-free namespace for future standardization (for example, by the taxonomies anticipated
+in <a href="https://github.com/w3c/webvtt/issues/512">WebVTT issue #512</a>). Reserved keys
 (such as "<code>lang</code>") have no hyphen and remain valid.</p>
 
 <p class="note">Keys are matched in a <a>case-sensitive</a> manner. Authors must use the
@@ -1715,7 +1715,7 @@ with all of the following properties:</p>
 this specification must use a <a>WebVTT header metadata value</a> that conforms to the
 constraints given for that key in [[#header-metadata]].</p>
 
-<p class="note">Note: Authors should avoid including multiple entries with same key in a single 
+<p class="note">Note: Authors should avoid including multiple entries with same key in a single
 <a>WebVTT file</a>. Only the first such entry will be used, and any subsequent entries ignored.</p>
 
 <p>A <dfn>WebVTT region definition block</dfn> consists of the following components, in the given
--- a/index.bs
+++ b/index.bs
@@ -1702,8 +1702,10 @@ with all of the following properties:</p>
 
 <p>A <a>WebVTT header metadata pair</a> whose <a>WebVTT header metadata key</a> is reserved by
 this specification must use a <a>WebVTT header metadata value</a> that conforms to the
-constraints given for that key in [[#header-metadata]]. A given reserved key must not appear
-more than once in a single <a>WebVTT file</a>.</p>
+constraints given for that key in [[#header-metadata]].</p>
+
+<p class="note">Note: Authors should avoid including multiple entries with same key in a single 
+<a>WebVTT file</a>. Only the first such entry will be used, and any subsequent entries ignored.</p>
 
 <p>A <dfn>WebVTT region definition block</dfn> consists of the following components, in the given
 order:</p>
--- a/index.bs
+++ b/index.bs
@@ -152,6 +152,7 @@ that captions an interview:</p>
 
 <pre>
 WEBVTT
+kind: subtitles
 
 00:11.000 --> 00:13.000
 &lt;v Roger Bingham>We are in New York City
@@ -225,6 +226,8 @@ one line except when a line break is definitely necessary.</p>
 
  <pre>
  WEBVTT
+ kind: captions
+ lang: en-UK
 
  00:01.000 --> 00:04.000
  Never drink liquid nitrogen.
@@ -342,7 +345,7 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
  <pre>
  WEBVTT
-
+ 
  STYLE
  ::cue {
    background-image: linear-gradient(to bottom, dimgray, lightgray);
@@ -435,6 +438,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
 
  test
  00:00.000 --> 00:02.000
@@ -478,6 +482,8 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
+ lang: fr-FR
 
  04:02.500 --> 04:05.000
  J'ai commencé le basket à l'âge de 13, 14 ans
@@ -497,6 +503,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
 
  00:00.000 --> 00:02.000
  &lt;v.first.loud Esme>It's a blue apple tree!
@@ -530,6 +537,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
 
  00:00:00.000 --> 00:00:04.000 position:10%,line-left align:left size:35%
  Where did he go?
@@ -584,6 +592,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
 
  REGION
  id:fred
@@ -639,6 +648,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
 
  00:01.000 --> 00:04.000
  Never drink liquid nitrogen.
@@ -658,6 +668,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: subtitles
 
  NOTE
  This file was written by Jill. I hope
@@ -696,6 +707,7 @@ might not be available.</p>
 
  <pre>
  WEBVTT
+ kind: chapters
 
  NOTE
  This is from a talk Silvia gave about WebVTT.
@@ -1310,6 +1322,7 @@ fragment:</p>
 
    <pre>
    WEBVTT
+   kind: subtitles
 
    00:00:07.000 --> 00:00:09.000
    What was his name again?
@@ -1335,6 +1348,7 @@ fragment:</p>
 
    <pre>
    WEBVTT
+   kind: subtitles
 
    00:00:04.000 --> 00:00:08.000
    <bdo dir=ltr>I've read the book &#x5E0;&#x5D5;&#x5D9;&#x5DC;&#x5E0;&#x5D3; 3 times!</bdo>
@@ -1360,7 +1374,7 @@ fragment:</p>
 
    <pre>
    WEBVTT
-
+   
    00:00:00.000 --> 00:00:05.000 align:start
    Hello!
    <bdo dir=ltr>&#x5E9;&#x5DC;&#x5D5;&#x5DD;!</bdo>
@@ -2420,6 +2434,7 @@ using only nested cues</dfn>:</p>
 
  <pre>
  WEBVTT
+ kind: chapters
 
  00:00.000 --> 01:24.000
  Introduction
@@ -2469,6 +2484,7 @@ using only nested cues</dfn>:</p>
 
  <pre>
  WEBVTT
+ kind: chapters
 
  00:00.000 --> 01:00.000
  The First Minute
--- a/index.bs
+++ b/index.bs
@@ -366,6 +366,63 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
 </div>
 
+<h3 id=introduction-header-metadata>WebVTT header metadata</h3>
+
+<p><i>This section is non-normative.</i></p>
+
+<p>A WebVTT file may include file-level metadata as key/value pairs on the lines immediately
+following the <code>WEBVTT</code> file header line. This is most useful for disambiguating the
+track <code>kind</code> and identifying the metadata schema in use, particularly for files
+delivered outside of an HTML context where a containing <code>&lt;track&gt;</code> element
+might not be available.</p>
+
+<div class="example">
+
+ <p>In this example, a captions track is identified with a kind, language, and human-readable
+ label.</p>
+
+ <pre>
+ WEBVTT
+ kind: captions
+ lang: es-MX
+ label: Español (SDH)
+
+ NOTE
+ Captions (SDH aka Subtitles for the Deaf and Hard-of-Hearing)
+ typically include spoken dialog as well as important audible
+ sounds such as "floor boards creak", "dogs barking", or in
+ this case, "[♫ música ♫]".
+
+ 1
+ 00:00:10.123 --> 00:00:15.432
+ ¡Hola! ¿Qué tál?
+
+ 2
+ 00:00:47.462 --> 00:01:04.028
+ [♫ música ♫]
+ </pre>
+
+</div>
+
+<div class="example">
+
+ <p>In this example, a descriptions track for blind or low-vision audiences is identified.
+ Descriptions are typically rendered as text-to-speech or braille rather than as on-screen
+ text.</p>
+
+ <pre>
+ WEBVTT
+ kind: descriptions
+ lang: en-US
+ label: English (AD)
+
+ 1
+ 00:00:10.123 --> 00:00:15.432
+ A young girl tiptoes down a dark hallway.
+ </pre>
+
+</div>
+
 <h3 id=introduction-other-features>Other caption and subtitling features</h3>
 
 <p><i>This section is non-normative.</i></p>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1455,6 +1515,65 @@ navigation tree.</p>
...
+
+<p><dfn>WebVTT header metadata</dfn> is an optional set of file-level key/value pairs declared at
+the start of a <a>WebVTT file</a>, immediately following the <code>WEBVTT</code> file header
+line and terminated by a blank line. It is represented as an ordered list of (key, value) string
+pairs.</p>
+
+<p>The keys listed below are reserved by this specification. Their values, when present, give
+the named property of the <a>WebVTT header metadata</a>; any pair whose key is not in this list
+is an <dfn>unrecognized header metadata pair</dfn> and is preserved for use by the consuming
+application.</p>
```
cookiecrook: Might need a more explicit reservation here. It may be that we may want to more strictly enforce custom keys include the hyphen as specified below, by only preserving those validly formatted keys in the "unrecognized" list. There could be a webcompat benefit to not allowing a random web app from using there own `foobar` key, but instead enforcing it as `x-foobar` or `more-specifically-namespaced-foobar`.

index.bs
```diff
@@ -1455,6 +1515,65 @@ navigation tree.</p>
...
+ </dd>
+
+ <dt><dfn lt="WebVTT header metadata type">type</dfn></dt>
+ <dd>
+  <p>An identifier for the schema or application that the <a lt="WebVTT metadata cue">metadata
+  cues</a> or <a lt="unrecognized header metadata pair">unrecognized header metadata pairs</a>
+  conform to. Disambiguates between different uses of <code>kind: metadata</code> and is the
+  primary mechanism for addressing the format ambiguity described in
+  <a href="https://github.com/w3c/webvtt/issues/511">WebVTT issue #511</a>.</p>
+ </dd>
```
cookiecrook, line 1582: I obviously can't leave this undefined and pointing to the issue. 

Instead, I might leave an editorial note here about the `type` key, and then address adding the key in #512 instead along with the PR for the "video.strobing.general-flash" type and point this to whatever registry list shakes out of that follow-on effort. Thoughts?

cookiecrook: There will of course be other types of metadata, but we need at least one to start the registry list.

index.bs
```diff
@@ -1501,6 +1625,83 @@ with the <a>MIME type</a> <code>text/vtt</code>. [[!RFC3629]]</p>
...
+ <li>It does not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D
+ HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
+</ul>
+
+<p>A <a>WebVTT header metadata key</a> must additionally satisfy one of the following:</p>
+
+<ul>
+ <li>It is one of the reserved key names defined for <a>WebVTT header metadata</a> in
+ [[#header-metadata]]: "<code>lang</code>", "<code>kind</code>", "<code>label</code>", or
+ "<code>type</code>".</li>
```
cookiecrook: Probably just reference the above list rather than duplicate it here.

index.bs
```diff
@@ -1501,6 +1625,83 @@ with the <a>MIME type</a> <code>text/vtt</code>. [[!RFC3629]]</p>
...
+with all of the following properties:</p>
+
+<ul>
+ <li>It contains no U+000A LINE FEED (LF) or U+000D CARRIAGE RETURN (CR) characters.</li>
+ <li>It contains none of the bidirectional formatting characters listed for
+ <a>WebVTT header metadata key</a>, above.</li>
+ <li>It does not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D
+ HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
+ <li>Its first and last characters are not U+0020 SPACE or U+0009 CHARACTER TABULATION (tab)
+ characters.</li>
```
cookiecrook, line 1713: Now wondering if this is strictly required. If space and tabs need trimming, would this lead to other trailing whitespace characters being used instead, and if so, can we get around without trimming anything other than the newline.

index.bs
```diff
@@ -1691,8 +1892,14 @@ separated from the next by a <a>WebVTT line terminator</a>. (In other words, any
...
-<p><a>WebVTT metadata text</a> cues are only useful for scripted applications (e.g. using the
-<code>metadata</code> <a>text track kind</a> in a HTML <a>text track</a>).</p>
+<p><a>WebVTT metadata text</a> cues are typically used by scripted applications (e.g. using the
+<code>metadata</code> <a>text track kind</a> in a HTML <a>text track</a>). Because the cue
+payload is otherwise an opaque string, authors of metadata files delivered outside of an HTML
+context should declare a <a lt="WebVTT header metadata kind">kind</a> of
+"<code>metadata</code>" together with a <a lt="WebVTT header metadata type">type</a>
+identifying the schema in use; see [[#header-metadata]]. Consumers that do not recognise the
+<a lt="WebVTT header metadata type">type</a> value should treat the cue payloads as opaque and
+must not present them as caption or subtitle text.</p>
```
cookiecrook: Nigel previously objected to the ambiguous "opaque" term, so I should remove it here.
```suggestion
<a lt="WebVTT header metadata type">type</a> value must not present the cue payloads as 
caption or subtitle text.</p>
```

index.bs
```diff
@@ -2809,13 +3035,121 @@ header</i> set, the user agent must run the following steps:</p>
...
+<h3 id=header-metadata-parsing algorithm>WebVTT header metadata parsing</h3>
+
+<p>When the algorithm in [[#file-parsing]] says to <dfn>collect WebVTT header metadata</dfn> from a
+string |input|, the user agent must run the following algorithm. The algorithm returns either a
+<a>WebVTT header metadata</a> object or null.</p>
+
+<ol algorithm="collect WebVTT header metadata">
+
+ <li><p>Let |metadata| be a new <a>WebVTT header metadata</a> object, with each of its reserved
+ keys ("<code>lang</code>", "<code>kind</code>", "<code>label</code>", "<code>type</code>")
```
cookiecrook: todo: reference the reserved key list, don't duplicate it here inline.

index.bs
```diff
@@ -2809,13 +3035,121 @@ header</i> set, the user agent must run the following steps:</p>
...
+   <li><p>Let |separator index| be the lowest index in |line| at which either a U+003A COLON
+   character or a U+003D EQUALS SIGN character appears.</p></li>
+
+   <li><p>Let |name| be the substring of |line| from the first character up to but not
+   including the character at |separator index|, with any trailing U+0020 SPACE and U+0009
+   CHARACTER TABULATION (tab) characters removed.</p></li>
+
+   <li><p>If |name| is not a <a>WebVTT header metadata key</a>, then jump to the step labeled
+   <i>next line</i>. (Invalid lines are ignored individually; they do not invalidate other
+   pairs.)</p></li>
```
cookiecrook:
```suggestion
   <li><p>If |name| is not a <a>WebVTT header metadata key</a>, then jump to the step labeled
   <i>next line</i>. Invalid lines are ignored individually; they do not invalidate other
   pairs.</p></li>
```

index.bs
cookiecrook: Discussed with @eric-carlson. Leaving as is for wider review.

himorin:
@cookiecrook 
pr-preview will remove lines in description **after** last of certain line, like `<!--` or `***`. I don't think automatic line by Claude will include such, but it is possible to be added to the bottom after pr-preview added (so, in result, order of `<description-text> <pr-preview lines> <claude-lines>` will be used, and latter two parts removed and new pr-preview lines added)
if you have any specific point to be considered, please file issue to https://github.com/tobie/pr-preview 

(note, not sure what the WG has decided on usage of AI tool to generate specification text...)

cookiecrook:
Thanks. I discovered my error in the description and corrected it. 

The context for others: I had commented wondering why PR Preview deleted the note about this being co-authored by GenAI, but deleted that comment once I discovered the reason and fixed the error. @himorin must've seen a record of the old comment in an older notification.

index.bs
```diff
@@ -1501,6 +1629,82 @@ with the <a>MIME type</a> <code>text/vtt</code>. [[!RFC3629]]</p>
...
+</ul>
+
+<p>A <a>WebVTT header metadata key</a> must additionally satisfy one of the following:</p>
+
+<ul>
+ <li>It is one of the keys reserved by this specification for <a>WebVTT header metadata</a>;
+ see [[#header-metadata]] for the current list of reserved keys.</li>
+ <li>It contains at least one U+002D HYPHEN-MINUS character, and its first character is not a
+ U+002D HYPHEN-MINUS character.</li>
+</ul>
```
nigelmegitt: This is confusingly presented. It means _either_ one _or_ the other condition must be met but that's very subtle - it took me several readings before I noticed the word "one" in line 1660 and realised that it means "exactly one".

There is more explanation in the note that follows, but it's too late (for this linear reader anyway!).

Suggest reformatting to be clearer that there are reserved keys, which are listed, and which contain no hyphen, and unreserved keys that can be defined by applications, which must contain at least one hyphen.

Then the note that follows could hopefully be removed or greatly reduced.

index.bs
```diff
@@ -1691,8 +1895,14 @@ separated from the next by a <a>WebVTT line terminator</a>. (In other words, any
 have two consecutive <a lt="WebVTT line terminator">WebVTT line terminators</a> and does not start
 or end with a <a>WebVTT line terminator</a>.)</p>
 
-<p><a>WebVTT metadata text</a> cues are only useful for scripted applications (e.g. using the
-<code>metadata</code> <a>text track kind</a> in a HTML <a>text track</a>).</p>
+<p><a>WebVTT metadata text</a> cues are typically used by scripted applications (e.g. using the
+<code>metadata</code> <a>text track kind</a> in a HTML <a>text track</a>). Authors of metadata
+files delivered outside of an HTML context should declare a
```
nigelmegitt: Is "should" correct here? It's definitely at least a "may", and not a "must".

index.bs
```diff
@@ -1691,8 +1895,14 @@ separated from the next by a <a>WebVTT line terminator</a>. (In other words, any
...
+<a lt="WebVTT header metadata kind">kind</a> of "<code>metadata</code>" together with a
+<a lt="WebVTT header metadata type">type</a> identifying the schema in use; see
+[[#header-metadata]]. Consumers that do not recognise the
+<a lt="WebVTT header metadata type">type</a> value must not present the cue payloads as
```
nigelmegitt: This seems to be asking a bit much - if the consumer doesn't recognise the WebVTT header metadata at all, then it has no way to reject presentation of the cue payloads based on the unrecognised "type" value.

index.bs
```diff
@@ -1501,6 +1629,82 @@ with the <a>MIME type</a> <code>text/vtt</code>. [[!RFC3629]]</p>
...
+ <li>It does not contain the substring "<code>--></code>" (U+002D HYPHEN-MINUS, U+002D
+ HYPHEN-MINUS, U+003E GREATER-THAN SIGN).</li>
+ <li>Its first and last characters are not U+0020 SPACE or U+0009 CHARACTER TABULATION (tab)
+ characters.</li>
+</ul>
+
+<p>A <a>WebVTT header metadata pair</a> whose <a>WebVTT header metadata key</a> is reserved by
+this specification must use a <a>WebVTT header metadata value</a> that conforms to the
+constraints given for that key in [[#header-metadata]]. A given reserved key must not appear
+more than once in a single <a>WebVTT file</a>.</p>
```
nigelmegitt: The parsing steps are clear that only the first pair for any given key will be used, and doesn't issue any parsing error or other error if this constraint is not met.

nigelmegitt (changes requested):
Another thought with no obvious place to comment on the changed files: in HTML there's a defined meaning to an attribute with no value, e.g. `<div id="..." someattr>` results in a DOM object that can be queried in javascript for the truthiness of `someattr`.

Would it make sense to match that here, and, in the absence of a `:` or `=` delimiter, if the entire line matches the requirements for a key, consider it to be a "true" attribute set on the file?

index.bs
cookiecrook: Is this just a statement, or are you asking for a change? If the latter, please clarify.

index.bs
cookiecrook:
```suggestion
files delivered outside of an HTML context may declare a
```

index.bs
cookiecrook: This is mainly intended to allow non-web consumers that **_do_** recognize the header metadata to know what types to **_not_** display as on-screen captions... Of course we can't update old clients where this is already a problem.

index.bs
cookiecrook: I will modify. Thanks.

index.bs
```diff
@@ -675,10 +732,13 @@ signifies the end of the WebVTT cue.</p>
...
 <div class="example">
 
- <p>In this example, a talk is split into each slide being a chapter.</p>
+ <p>In this example, topics mentioned in a talk are provided as URLs for reference. The
+ <code>kind: metadata</code> <a>WebVTT header metadata</a> pair signals to consumers that the
+ cue payloads should be processed by a script rather than displayed as text.</p>
 
  <pre>
  WEBVTT
+ kind: metadata
```
gkatsev, line 752: thought(non-blocking): I wonder whether it would be helpful to add these headers to other examples?

gkatsev:
> Another thought with no obvious place to comment on the changed files: in HTML there's a defined meaning to an attribute with no value, e.g. `<div id="..." someattr>` results in a DOM object that can be queried in javascript for the truthiness of `someattr`.
> 
> Would it make sense to match that here, and, in the absence of a `:` or `=` delimiter, if the entire line matches the requirements for a key, consider it to be a "true" attribute set on the file?

I wonder whether we should allow it at the moment or to keep the scope smaller disallow it for now until this is asked for. I'd lean towards disallow at the moment, which I believe is currently covered by the parsing logic.

index.bs
gkatsev: While I understand why it's added, I wonder if it's a bit too restrictive without a clear benefit. Metadata tracks are already generally expected to be used in a scripting context and not shown to users.
This is definitely an edge case, but what should a vendor do is the following situation?
- someone sets a kind:metadata and a type but then in the html app that track's mode is set to `showing`?
it might be ok to display it since it was explicitly set but a this line says that the it must not be displayed.

index.bs
gkatsev: I don't remember if we talked about this in the other PR, I didn't find a direct answer from looking briefly.

gkatsev (commented):
Thanks! Seems like we're in a pretty good place with this PR now.

gkatsev:
one thing I noticed that isn't relevant to the contents of the PR is that in the diff preview extra whitespace is collapsed in the examples. So, everything is bunched up together.
image
Seems to be fine in the regular preview and the regular deploy.

index.bs
nigelmegitt: I think a change is needed to clarify what happens if this "must" constraint is not met, i.e. what happens if the reserved key is used more than once? Perhaps "must" is not the right term here.

chrisn: The algorithm steps will mean the first use of a key is added to the WebVTT header metadata object and any subsequent ones are ignored ("do nothing" in the algorithm). So the only change I'd suggest here is to rephrase:

> A given reserved key must not appear more than once in a single <a>WebVTT file</a>.

as an explanatory guidance for authors, in a Note, e.g:

> Note: Authors should avoid including multiple entries with same key in a single WebVTT file. Only the first such entry will be used, and any subsequent entries ignored.

index.bs
cookiecrook: > Consumers that do not recognise the `type` value must not present the cue payloads as caption or subtitle text.

Options for resolution seem to be:
1. cut this sentence entirely
2. change the `must not` to a `should not`
3. change the `must not` to a `may not` (or would this be `might not`?) 

I'm going with 1 (cut the sentence) unless I hear otherwise.

index.bs
cookiecrook: YEs, I had some in the ATTRIBUTES block PR. I will re-insert.

cookiecrook: https://github.com/w3c/webvtt/pull/548/changes/7ee1c379e5fb7d78ea87aa50fbb48c42183ad082 adds `kind: subtitles` to most examples except the text formatting examples because those got too noisy. I added `lang:` to the examples where language was most relevant. Please review. Feedback especially requested on whether I chose correctly on all of the `kind: chapters` examples, and if you think any of the `subtitles` should be `captions`. Thanks.

index.bs
cookiecrook: Resolved in https://github.com/w3c/webvtt/pull/548/changes/2a0f1a3ae2dc14f32966c86d13c42a8b77df68a3

index.bs
cookiecrook: Resolved in https://github.com/w3c/webvtt/pull/548/changes/fa2260fd8b2016befbc9ab6fe01e46fa76ff52f8

index.bs
```diff
@@ -342,7 +345,7 @@ CSS comment (e.g. <code>/**/</code>).</p>
 
  <pre>
  WEBVTT
-
+
```
cookiecrook:
```suggestion

```

index.bs
```diff
@@ -225,6 +226,8 @@ one line except when a line break is definitely necessary.</p>
 
  <pre>
  WEBVTT
+ kind: captions
+ lang: en-UK
```
cookiecrook: lang not relevant here.
```suggestion
```

index.bs
cookiecrook: Note: I resolved a couple nits (whitespace diffs and the irrelevant en-UK line) in a later commit.

index.bs
```diff
@@ -1300,7 +1374,7 @@ fragment:</p>
 
    <pre>
    WEBVTT
-
+
```
cookiecrook:
```suggestion

```

cookiecrook:
@nigelmegitt wrote:
> > Another thought with no obvious place to comment on the changed files: in HTML there's a defined meaning to an attribute with no value, e.g. `<div id="..." someattr>` results in a DOM object that can be queried in javascript for the truthiness of `someattr`.
> > Would it make sense to match that here, and, in the absence of a `:` or `=` delimiter, if the entire line matches the requirements for a key, consider it to be a "true" attribute set on the file?

@gkatsev replied
> I wonder whether we should allow it at the moment or to keep the scope smaller disallow it for now until this is asked for. I'd lean towards disallow at the moment, which I believe is currently covered by the parsing logic.

I think Nigel is referring to [Boolean attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes). If so, those are all implicitly false, but scope limited to specific Element definitions. For example, `required` is allowed on any form input element, but the `checked` attribute is only allowed on checkbox. 

Without a way to limit the scope of which implicit Boolean metadata keys exist (like IDL provides for HTML elements), we'd be assigning permanent "falsy" values to every undefined key going forward. I agree with Gary that I'd prefer to keep the scope smaller now until a need is presented later.

index.bs
cookiecrook: I struggled with this a bit… Here's the current version, which is now part of the previous list, rather than a separate "additionally" list.

> - [prior requirements for a key]
> - It must _either_ match a reserved key, _or_ must match the additional hyphenated formatting requirements of a custom key:
>   - **reserved key**: Matches one of the keys reserved by this specification for WebVTT header metadata; see [[#header-metadata]] for the current list of reserved keys.
>   - **custom key**: contains at least one U+002D HYPHEN-MINUS character, and its first character is not a U+002D HYPHEN-MINUS character.

cookiecrook: @nigelmegitt I think this is the last issue blocking review. All other review feedback has been resolved. Please let me know if you have additional comments. Thanks!

cookiecrook:
@nigelmegitt I just fixed a few errant whitespace diffs I somehow missed previously. I think this is the only remaining discussion issue: a rewrote of some prose you correctly noted was confusingly phrased. Thanks for your review.

- https://github.com/w3c/webvtt/pull/548#discussion_r3502716290

